PR #77, opened 2015-07-15
Tweaks to get Summa to compile on *buntu machines

naught101:
Default settings for a vanilla debian/ubuntu style install (only requries the user to input "gfortran" as the FC, without specifying a version).

Also, a note in the readme about the atlas requirement.


bartnijssen:
@naught101 : Can you re-issue the PR against `UCAR:develop`?

naught101:
Done: #78

PR #97, opened 2015-11-21
implement double exponential option for root density

martynpclark:
This feature extends options to calculate root density.

Most changes are in the subroutine rootDensty within var_derive.f90, where an option is added for the double exponential rooting profile introduced by Jackson 1996 and formulated for land models by Zeng 1998.

Other changes are:
(1) get_ixname.f90: Add global variables for root density
(2) mDecisions.f90: Add decisions for root density.

Notes:
(1) The branch feature/stomatalConductance was merged prior to making the pull request. Please either review and merge feature/stomatalConductance first, or compare against feature/stomatalConductance.
(2) The code was run for all test cases, providing results within machine precision of the simulation results from the master branch.


bartnijssen:
a) travis checks fail, which means it is not compiling, see travis for details: https://travis-ci.org/NCAR/summa/builds/93988640

b) please go ahead and merge `NCAR:develop` into this branch, which should get rid of your merge conflicts. I think it will also reduce the size of the PR, which has some elements of the canopy resistance work in it.

PR #99, opened 2015-11-24
Update summa_and_git_howto (again)

naddor:
Based on discussions with Ethan and Andy, and on my own experience working with summa_tools, it seems that -track can be removed.


bartnijssen:
There are definitely advantages to use tracking: http://git-scm.com/book/en/v2/Git-Branching-Remote-Branches#Tracking-Branches , but if I understand it correctly the advantage may be at the level of individual branches rather than an entire remote repo.

PR #114, opened 2016-02-03
Adding ONLY the files for git-lfs for GRU commit.

anewman89:
This is a pull request to test git-lfs for SUMMA.  I've added only one file supported by git-lfs.  I'd like to then fetch upstream and merge to my local SUMMA and see if I can push back to my fork.  If I can't I think we're still going to have problems with git-lfs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*tgz filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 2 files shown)


bartnijssen:
@anewman89 : Do we still want to merge this?

anewman89:
I think we can kill this merge request.  Since we're likely going to use the submodule route, I don't hitnk this isn't relevant anymore unless someone else says otherwise.

anewman89:
Can I just close this request then?

PR #117, opened 2016-03-26
Feature/netcdf input

martynpclark:
Merges Andy Newman's netcdf routines and cleans them up a little.

This PR matches output from feature/refactorNumerix to within machine precision.

NOTE: This PR should be processed AFTER feature/mergeGRU

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -150,7 +150,10 @@ DATAMS = $(patsubst %, $(DSHARE_DIR)/%, $(SUMMA_DATAMS))
 
 # utility modules
 SUMMA_UTILMS= \
-		time_utils.f90
+		get_ixname.f90 \
+		time_utils.f90 \
+		ascii_util.f90 \
+		mDecisions.f90
 UTILMS = $(patsubst %, $(ENGINE_DIR)/%, $(SUMMA_UTILMS))
 
 # Model guts
@@ -176,11 +179,7 @@ SOLVER = $(patsubst %, $(ENGINE_DIR)/%, $(SUMMA_SOLVER))
 # Define routines for SUMMA preliminaries
 SUMMA_PRELIM= \
 		conv_funcs.f90 \
-		netcdf_util.f90 \
-		ascii_util.f90 \
 		sunGeomtry.f90 \
-		get_ixname.f90 \
-		mDecisions.f90 \
 		snow_utils.f90 \
 		soil_utils.f90 \
 		updatState.f90 \
@@ -228,6 +227,7 @@ SUMMA_MSOLVE = \
 
 # Define NetCDF routines
 SUMMA_NETCDF = \
+		netcdf_util.f90 \
 		def_output.f90 \
 		modelwrite.f90
 NETCDF = $(patsubst %, $(NETCDF_DIR)/%, $(SUMMA_NETCDF))
@@ -236,7 +236,7 @@ NETCDF = $(patsubst %, $(NETCDF_DIR)/%, $(SUMMA_NETCDF))
 COMM_ALL = $(NRUTIL) $(NRPROC) $(HOOKUP) $(DATAMS) $(UTILMS)
 
 # ... stitch together SUMMA programs
-SUMMA_ALL = $(PRELIM) $(MODRUN) $(SOLVER) $(NETCDF)
+SUMMA_ALL = $(NETCDF) $(PRELIM) $(MODRUN) $(SOLVER)
 
 # Define the driver routine
 SUMMA_DRIVER= \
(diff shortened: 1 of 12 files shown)


build/source/engine/time_utils.f90
```diff
@@ -174,5 +175,92 @@ subroutine compjulday(iyyy,mm,id,ih,imin,dsec,&  ! input
...
+ integer(i4b), intent(out)     :: mm           ! month
+ integer(i4b), intent(out)     :: id           ! day
+ integer(i4b), intent(out)     :: ih           ! hour
+ integer(i4b), intent(out)     :: imin         ! minute
+ real(dp),     intent(out)     :: dsec         ! seconds
+ integer(i4b), intent(out)     :: err          ! error code
+ character(*), intent(out)     :: message      ! error message
+
+ ! local parameters
+ integer(i4b),parameter       :: y = 4716
```
greyNearing: Comments on these guys would be nice.

build/source/engine/time_utils.f90
```diff
@@ -174,5 +175,92 @@ subroutine compjulday(iyyy,mm,id,ih,imin,dsec,&  ! input
...
+ integer(i4b),parameter       :: n = 12
+ integer(i4b),parameter       :: r = 4
+ integer(i4b),parameter       :: p = 1461
+ integer(i4b),parameter       :: v = 3
+ integer(i4b),parameter       :: u = 5
+ integer(i4b),parameter       :: s = 153
+ integer(i4b),parameter       :: w = 2
+ integer(i4b),parameter       :: b = 274277
+ integer(i4b),parameter       :: c = -38
+ real(dp),parameter           :: hr_per_day = 24.0_dp
```
greyNearing: Physical parameters, technically. I might move these to globalData.

build/source/engine/time_utils.f90
```diff
@@ -174,5 +175,92 @@ subroutine compjulday(iyyy,mm,id,ih,imin,dsec,&  ! input
...
+ real(dp)              :: frac_day  ! fractional day 
+ real(dp)              :: remainder ! remainder of modulus operation
+
+ ! initialize errors
+ err=0; message="compcalday"
+ if(julday<=0)then;err=10;message=trim(message)//"no negative julian days/"; return; endif
+
+ ! step 1
+ step_1a = 4*julday+b
+ step_1b = step_1a/146097
```
greyNearing: What are these constants?

build/Makefile
```diff
@@ -237,7 +237,7 @@ NETCDF = $(patsubst %, $(NETCDF_DIR)/%, $(SUMMA_NETCDF))
 COMM_ALL = $(NRUTIL) $(NRPROC) $(HOOKUP) $(DATAMS) $(UTILMS)
 
 # ... stitch together SUMMA programs
-SUMMA_ALL = $(PRELIM) $(MODRUN) $(SOLVER) $(NETCDF)
+SUMMA_ALL = $(NETCDF) $(PRELIM) $(MODRUN) $(SOLVER)
```
greyNearing: Doesn't `def_output.f90` use `data_types`?

build/source/engine/time_utils.f90
bartnijssen: Can we avoid hard-coded literals like this entirely (at least here). See also issue #118

martynpclark: This is a subroutine that Andy Neman extracted from wikipedia. The source is given in
`! source: https://en.wikipedia.org/wiki/Julian_day#Julian_or_Gregorian_calendar_from_Julian_day_number`
at the top of the subroutine.

To be honest, I don't think users need to understand why this works. However, we should clean up out time utilities -- and make things like this less cryptic.

I've added a note to issue #124 to improve time utilities.

build/source/engine/time_utils.f90
martynpclark: See comment above. Added to issue #124

build/source/engine/time_utils.f90
martynpclark: See comment above. Added to issue #124

build/source/engine/time_utils.f90
bartnijssen: I now recognize that these may be part of different calendar calculations. I would probably leave them (redefining as parameters is not really going to help as most are used only once). Provide the source of these calculations if possible. In VIC we have a similar routine, but also provide the references. If the values are only used once and only used here, it may even be clearer to just leave the literals in the equations. See for example https://github.com/UW-Hydro/VIC/blob/develop/vic/drivers/shared_all/src/vic_time.c

build/Makefile
martynpclark: Correct.

However, `data_types` are compiled in `COMM_ALL` above.

PR #134, opened 2016-04-01
Martynpclark bug fix/convergence problems

bartnijssen:


bartnijssen:
Not quite what we wanted

PR #136, opened 2016-04-08
Incorrect function evaluation in lineSearch

martynpclark:
The previous code computed the function norm2(rVec,rVec)/2, which was actually inconsistent with the gradient calculations used in lineSearch. The correct function is dot_product(rVec,rVec)/2, and the code has been updated.

The code has been checked with all test cases, with correlations between the bugFix/wrongFunction branch and the bugFix/develop branch >0.99. We do not match numerical precision because the numerical solution has changed.

Spot checks indicate a 10% speed-up.

### DIFF
--- a/build/source/engine/systemSolv.f90
+++ b/build/source/engine/systemSolv.f90
@@ -666,9 +666,6 @@ contains
  if(err/=0)then; message=trim(message)//trim(cmessage); return; endif  ! (check for errors)
 
  ! compute the function evaluation
- !fOld=0.5_dp*norm2(real(rVec, dp)/fScale)  ! NOTE: norm2 = sqrt(sum((rVec/fScale)**2._dp))
-
-
  rVecScaled = real(rVec, dp)/fScale
  fOld=0.5_dp*dot_product(rVecScaled,rVecScaled) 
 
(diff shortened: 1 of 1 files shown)


build/source/engine/systemSolv.f90
```diff
@@ -665,7 +666,11 @@ fScale,                  & ! intent(out):   function scaling vector (mixed units
  if(err/=0)then; message=trim(message)//trim(cmessage); return; endif  ! (check for errors)
 
  ! compute the function evaluation
- fOld=0.5_dp*norm2(real(rVec, dp)/fScale)  ! NOTE: norm2 = sqrt(sum((rVec/fScale)**2._dp))
+ !fOld=0.5_dp*norm2(real(rVec, dp)/fScale)  ! NOTE: norm2 = sqrt(sum((rVec/fScale)**2._dp))
```
bartnijssen: delete commented out lines

build/source/engine/systemSolv.f90
```diff
@@ -2770,7 +2775,10 @@ f,                       & ! intent(out): new function value (mixed units)
    if(err/=0)then; message=trim(message)//trim(cmessage); return; endif  ! (check for errors)
 
    ! compute the function evaluation
-   f=0.5_dp*norm2(real(rVec, dp)/fScale)  ! NOTE: norm2 = sqrt(sum((rVec/fScale)**2._dp))
+   !f=0.5_dp*norm2(real(rVec, dp)/fScale)  ! NOTE: norm2 = sqrt(sum((rVec/fScale)**2._dp))
```
bartnijssen: delete commented out lines

PR #137, opened 2016-04-09
Feature/refactor system solv

martynpclark:
This branch splits the numerical routine `systemSolv.f90` into several additional routines. Specifically, we now have
* `eval8summa.f90` which computes the fluxes and residuals for a given state vector. The routine `eval8summa.f90` calls `computFlux.f90` and `computResid.f90` to compute the model fluxes and residuals respectively.
* `summaSolv.f90` computes the new state vector for a current iteration. It does this by first computing the Jacobian matrix using `computJacob.f90` based on the derivatives computed in the previous call to `eval8summa.f90` (either at the start of `systemSolv.f90` or from a previous iteration), then computing the iteration increment in `lapackSolv`, and then by evaluating (and, if necessary, refining) the new state vector in `lineSearchRefinement` or `trustRegionRefinement`.
* The evaluation of the state vector in `lineSearchRefinement` or `trustRegionRefinement` calls `eval8summa.f90`, which computes the fluxes and derivatives used in subsequent calls to `summaSolv.f90`.

This branch has been run through all test cases, showing
* The simulations are almost identical to the develop branch (correlations greater than 0.99), though results do not match to within machine precision. We expect bit-for-bit matching since the only change has been reorganization of the code.
* The simulations are slower, caused by more flux evaluations.

While these results are sub-optimal, I still make this pull request in order to avoid excessive divergence among the different branches. I will make a new branch to address these remaining issues.


build/source/dshare/var_lookup.f90
```diff
@@ -24,8 +24,9 @@ MODULE var_lookup
  implicit none
  private
  ! local variables
- integer(i4b),parameter     :: imiss = -999             ! missing value: used to initialize named variables
- integer(i4b),parameter     :: iLength=storage_size(1)  ! size of an integer
+ integer(i4b),parameter     :: imiss = -999                 ! missing value: used to initialize named variables
```
greyNearing: Use integerMissing from globalData.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,851 @@
...
+ integer(i4b),intent(out)        :: err                       ! error code
+ character(*),intent(out)        :: message                   ! error message
+ ! ---------------------------------------------------------------------------------------
+ ! * local variables
+ ! ---------------------------------------------------------------------------------------
+ integer(i4b)                    :: local_ixGroundwater       ! local index for groundwater representation
+ integer(i4b)                    :: iSoil                     ! index of soil layer
+ integer(i4b)                    :: iLayer                    ! index of model layers
+ real(dp)                        :: theta                     ! liquid water equivalent of total water (liquid plus ice)
+ real(dp),parameter              :: canopyTempMax=500._dp     ! expected maximum value for the canopy temperature (K)
```
greyNearing: Physical constant belongs in an external module.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,851 @@
...
+ freeDrainage,                    & ! free drainage
+ liquidFlux,                      & ! liquid water flux
+ zeroFlux                           ! zero flux
+
+implicit none
+private
+public::computFlux
+public::soilCmpres
+! control parameters
+real(dp),parameter  :: valueMissing=-9999._dp     ! missing value
```
greyNearing: Should make all missings global, since these are essentially flags, their job is communication, and therefore, these are truly global variable.s

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,851 @@
...
+real(dp),parameter  :: verySmall=tiny(1.0_dp)     ! a very small number
```
greyNearing: Also probably program variables - make global.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,851 @@
...
+real(dp),parameter  :: veryBig=1.e+20_dp          ! a very big number
+real(dp),parameter  :: dx = 1.e-8_dp              ! finite difference increment
```
greyNearing: Might be good to give user control over the various optimization threshold parameters. I'm sure there are others.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,851 @@
...
+   dTheta_dTkCanopy = dFracLiq_dTk(scalarCanopyTempTrial,snowfrz_scale)*theta   ! K-1
+   dCanLiq_dTcanopy = dTheta_dTkCanopy*iden_water*canopyDepth                   ! kg m-2 K-1
+  else
+   dTheta_dTkCanopy = 0._dp
+   dCanLiq_dTcanopy = 0._dp
+  endif
+ endif
+
+ ! * snow+soil domain: compute derivative of volumetric liquid water content w.r.t. temperature (K-1)
+ do iLayer=1,nLayers  ! loop through all snow and soil layers
```
luckymichael: Faster through two loops (snow and soil) without select case?

build/source/engine/computFlux.f90
luckymichael: What are the difference between realMissing and valueMissing?

valueMissing is not used.

build/source/dshare/var_lookup.f90
martynpclark: This is tricky. The routine `globalData.f90` actually uses the module `var_lookup`, and it would require some reorganization to define missing values that can be used in `var_lookup`.

Ultimately we may want two forms of global data: (1) global constants (e.g., put global variables in `multiconst.f90`); and (2) global data structures (e.g., the metadata structures). This would enable us to use `integerMissing` in `var_lookup`

This extent of reorganization goes beyond what is attempted here, and is left for a future refactor.

build/source/engine/computFlux.f90
martynpclark: True. Physical constants do belong in an external module.

Suggest we think of the different types of global constants. These can include:
* Physical constants
* Limits used for error checking (as is done here)
* Flags (e.g., missing values)

Once we have the different categories defined we can reorganize the code to clarify the role of the different constants (and avoid the circuitous problems identified in the comment above). 

This reorganization should be the topic of a future PR, and I've opened up an issue to address this (issue #140).

build/source/engine/computFlux.f90
martynpclark: The variable `valueMissing` has been deleted. It was replaced with `realMissing`

build/source/engine/computFlux.f90
martynpclark: Done

build/source/engine/computFlux.f90
martynpclark: The user does have control over most algorithmic control parameters. There are some -- like the finite difference increment -- where we may not want to give the user control as this can cause problems if set incorrectly.

build/source/engine/computFlux.f90
martynpclark: Good point. There is likely substantial speedup possible through identifying all such instances. It would be good if these are all addressed together and systematically. I've opened up a new issue (issue #142) to address this directly.

build/source/dshare/globalData.f90
```diff
@@ -100,6 +112,24 @@ MODULE globalData
...
+ integer(i4b),parameter,public               :: ixSub1=kl+5             ! index for the 1st sub-diagonal band
+ integer(i4b),parameter,public               :: ixSub2=kl+6             ! index for the 2nd sub-diagonal band
+ integer(i4b),parameter,public               :: ixSub3=kl+7             ! index for the 3rd sub-diagonal band
+ integer(i4b),parameter,public               :: ixSub4=kl+8             ! index for the 3rd sub-diagonal band
+ integer(i4b),parameter,public               :: nBands=2*kl+ku+1        ! length of the leading dimension of the band diagonal matrix
+ integer(i4b),parameter,public               :: ixFullMatrix=1001       ! named variable for the full Jacobian matrix
+ integer(i4b),parameter,public               :: ixBandMatrix=1002       ! named variable for the band diagonal matrix
+ integer(i4b),parameter,public               :: iJac1=1                 ! first layer of the Jacobian to print
+ integer(i4b),parameter,public               :: iJac2=10                ! last layer of the Jacobian to print
+
```
bartnijssen: Can you provide some comments about the above indexing scheme? If they are all just indexes, why are they offset by `kl`? As is, this is difficult to understand. Why would the `ixSupx` indices not be offset by `ku` and why is the total number of bands `nBands` equal tp `2*kl+ku+1` rather than `kl+ku+1`?

Is something special about the 10th layer of the Jacobian?

build/source/dshare/popMetadat.f90
```diff
@@ -511,22 +514,22 @@ subroutine popMetadat(err,message)
...
+ deriv_meta(iLookDERIV%dCompress_dPsi)                = var_info('dCompress_dPsi'               , 'derivative in compressibility w.r.t matric head'                      , 'm-1'            , 'midSoil', .false.)
+ deriv_meta(iLookDERIV%mLayerdTheta_dPsi)             = var_info('mLayerdTheta_dPsi'            , 'derivative in the soil water characteristic w.r.t. psi'               , 'm-1'            , 'midSoil', .false.)
+ deriv_meta(iLookDERIV%mLayerdPsi_dTheta)             = var_info('mLayerdPsi_dTheta'            , 'derivative in the soil water characteristic w.r.t. theta'             , 'm'              , 'midSoil', .false.)
+ deriv_meta(iLookDERIV%dq_dHydStateAbove)             = var_info('dq_dHydStateAbove'            , 'change in flux at layer interfaces w.r.t. states in the layer above'  , 'unknown'        , 'ifcSoil', .false.)
+ deriv_meta(iLookDERIV%dq_dHydStateBelow)             = var_info('dq_dHydStateBelow'            , 'change in flux at layer interfaces w.r.t. states in the layer below'  , 'unknown'        , 'ifcSoil', .false.)
```
bartnijssen: `unknown` or dependent on the formulation? Fine as is, just for my own edification.

build/source/engine/allocspace.f90
```diff
@@ -210,33 +210,49 @@ subroutine allocGlobal(metaStruct,dataStruct,err,message)
...
+   associate(&
+   nSnow => gru_struc(iGRU)%hruInfo(iHRU)%nSnow, & ! number of snow layers for each HRU
+   nSoil => gru_struc(iGRU)%hruInfo(iHRU)%nSoil  ) ! number of soil layers for each HRU
+
+   ! allocate space for structures WITH an HRU dimension
+   select type(dataStruct)
+    type is (gru_hru_int);       call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow,nSoil,err,cmessage); spatial=.true.
+    type is (gru_hru_intVec);    call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow,nSoil,err,cmessage); spatial=.true.
+    type is (gru_hru_double);    call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow,nSoil,err,cmessage); spatial=.true.
+    type is (gru_hru_doubleVec); call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow,nSoil,err,cmessage); spatial=.true.
```
bartnijssen: Since all four calls are identical, do you still needs the `select()` statement here?

build/source/engine/allocspace.f90
```diff
@@ -210,33 +210,49 @@ subroutine allocGlobal(metaStruct,dataStruct,err,message)
...
-   type is (gru_hru_intVec);    do iHRU=1,nHRU; call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow(iHRU),nSoil(iHRU),err,cmessage); if(err/=0)exit; end do
-   type is (gru_hru_double);    do iHRU=1,nHRU; call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow(iHRU),nSoil(iHRU),err,cmessage); if(err/=0)exit; end do
-   type is (gru_hru_doubleVec); do iHRU=1,nHRU; call allocLocal(metaStruct,dataStruct%gru(iGRU)%hru(iHRU),nSnow(iHRU),nSoil(iHRU),err,cmessage); if(err/=0)exit; end do
-   ! structures without an HRU dimension
-   type is (gru_double);    call allocLocal(metaStruct,dataStruct%gru(iGRU),nSnow=0,nSoil=0,err=err,message=cmessage)
-   type is (gru_doubleVec); call allocLocal(metaStruct,dataStruct%gru(iGRU),nSnow=0,nSoil=0,err=err,message=cmessage)
-   class default; spatial=.false.
+   type is (gru_double);    call allocLocal(metaStruct,dataStruct%gru(iGRU),nSnow=0,nSoil=0,err=err,message=cmessage); spatial=.true.
+   type is (gru_doubleVec); call allocLocal(metaStruct,dataStruct%gru(iGRU),nSnow=0,nSoil=0,err=err,message=cmessage); spatial=.true.
```
bartnijssen: Same as above - no `select` required.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+  err=20; return
+ endif
+
+ ! *****
+ ! * COMPUTE DERIVATIVES ASSOCIATED WITH MELT-FREEZE...
+ ! ****************************************************
+
+ ! * vegetation domain: compute derivative of volumetric liquid water content w.r.t. temperature (K-1)
+ if(computeVegFlux)then
+  if(scalarCanopyIceTrial > verySmall)then
```
bartnijssen: Be consistent in spacing around '>' (below you have no spacing)

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+
+ end do  ! (looping through soil layers)
+
+ ! initialize liquid water fluxes throughout the snow and soil domains
+ ! NOTE: used in the energy routines, which is called before the hydrology routines
+ if(firstFluxCall)then
+  if(nSnow > 0)&
+  iLayerLiqFluxSnow(0:nSnow) = 0._dp
+  iLayerLiqFluxSoil(0:nSoil) = 0._dp
+ endif
```
bartnijssen: Is there a missing `endif`? Or so you get around that with the line continuation. If so, may be clearer to write:

```Fortran
if(firstFluxCall)then
 if(nSnow > 0) iLayerLiqFluxSnow(0:nSnow) = 0._dp
 iLayerLiqFluxSoil(0:nSoil) = 0._dp
endif
```

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+  else
+   scalarSoilControl = 0._dp  ! (scalarRainPlusMelt exceeds maximum infiltration rate
+  endif
+ else
+  ! * case of infiltration into snow
+  scalarSoilControl = 1._dp
+ endif
+
+ ! expand derivatives to the total water matric potential
+ dq_dHydStateAbove(1:nSoil)   = dq_dHydStateAbove(1:nSoil)  *dPsiLiq_dPsi0(1:nSoil)
```
bartnijssen: Explain in comment why the arrays are offset

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+ dq_dHydStateBelow(0:nSoil-1) = dq_dHydStateBelow(0:nSoil-1)*dPsiLiq_dPsi0(1:nSoil)
+
+ ! *****
+ ! * CALCULATE THE GROUNDWATER FLOW...
+ ! ************************************
+
+ ! set baseflow fluxes to zero if the baseflow routine is not used
+ if(local_ixGroundwater/=qbaseTopmodel)then
```
bartnijssen: As we get more parameterizations the easier way to do this may be to always initialize the values to some default values and then overwrite them with the appropriate values based on the selected method. I.e. avoid having to loop through cases that say `(selectedMethod /= methodX)` and only have cases that say `(selectedMethod == methodX)`. Otherwise it will become harder to maintain when we add more methods.

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+                  mLayerBaseflow,                          & ! intent(out): baseflow from each soil layer (m s-1)
+                  dBaseflow_dMatric,                       & ! intent(out): derivative in baseflow w.r.t. matric head (s-1)
+                  err,cmessage)                              ! intent(out): error control
+  if(err/=0)then; message=trim(message)//trim(cmessage); return; endif
+ endif  ! computing baseflow flux
+
+ ! compute total baseflow from the soil zone (needed for mass balance checks)
+ scalarSoilBaseflow = sum(mLayerBaseflow)
+
+
```
bartnijssen: remove extra whitespace

build/source/engine/computFlux.f90
```diff
@@ -0,0 +1,854 @@
...
+ ! *****
+ ! (7) CALCUALTE FLUXES FOR THE DEEP AQUIFER...
+ ! ********************************************
+
+ ! identify modeling decision
+ if(local_ixGroundwater==bigBucket)then
+  ! deep aquifer is not yet transfered from old code structure
+  message=trim(message)//'bigBucket groundwater parameterization is not yet transfered from old code structure'
```
bartnijssen: `transfered` should read `transferred`

build/source/engine/computJacob.f90
```diff
@@ -0,0 +1,527 @@
...
+   ! -----
+   ! * energy fluxes for the snow-soil domain...
+   ! -------------------------------------------
+   do iLayer=1,nLayers  ! loop through layers in the snow-soil domain
+    ! - define layer indices
+    jLayer = ixSnowSoilNrg(iLayer)
+    ! - compute the Jacobian
+    aJac(jLayer,jLayer)   = (dt/mLayerDepth(iLayer))*(-dNrgFlux_dTempBelow(iLayer-1) + dNrgFlux_dTempAbove(iLayer)) + dMat(jLayer)
+    if(iLayer > 1)       aJac(jLayer-nVarSnowSoil,jLayer) = (dt/mLayerDepth(iLayer-1))*( dNrgFlux_dTempBelow(iLayer-1) )
+    if(iLayer < nLayers) aJac(jLayer+nVarSnowSoil,jLayer) = (dt/mLayerDepth(iLayer+1))*(-dNrgFlux_dTempAbove(iLayer  ) )
```
bartnijssen: Just making sure the logic here is intentional:
* First `if` statements executed for `iLayer > 1`
* Second `if` for `iLayers < nLayers`
So for layers in the range `[2, nLayers-1]` both statements are executed

build/source/engine/computResid.f90
```diff
@@ -0,0 +1,201 @@
...
+
+ ! compute energy associated with melt freeze for the vegetation canopy
+ if(computeVegFlux)then
+  rAdd(ixVegNrg) = rAdd(ixVegNrg) + LH_fus*(scalarCanopyIceTrial - scalarCanopyIce)/canopyDepth   ! energy associated with melt/freeze (J m-3)
+ endif
+
+ ! compute energy associated with melt/freeze for snow
+ ! NOTE: allow expansion of ice during melt-freeze
+ if(nSnow>0)&
+ rAdd(ixSnowOnlyNrg) = rAdd(ixSnowOnlyNrg) + LH_fus*iden_ice*(mLayerVolFracIceTrial(1:nSnow) - mLayerVolFracIce(1:nSnow))       ! energy associated with melt/freeze (J m-3)
```
bartnijssen: If you are splitting them over two layers anyway, this is much easier to read without the line continuation:

```Fortran
if()
...
endif
```

build/source/engine/matrixOper.f90
```diff
@@ -0,0 +1,231 @@
...
+ ! initialize error control
+ err=0; message='scaleMatrices/'
+
+ ! select the type of matrix
+ select case(ixMatrix)
+
+  ! * full matrix
+  case(ixFullMatrix)
+
+   ! scale the by both the function and variable scaling factors
```
bartnijssen: Fix comment

bartnijssen:
@martynpclark : I'll admit that I started glazing over towards the end. As always - if you decide not to fix an issue just state so or if it is warranted open an issue and refer to that.

Let me know when you have gone through it and I can go ahead and merge.

build/source/dshare/globalData.f90
martynpclark: Done. This indexing scheme provides the matrix structure expected by lapack. Specifically, lapack requires `kl` extra rows for additional storage. Consequently, all indices are offset by `kl` and the total number of bands for storage is `2*kl+ku+1` instead of `kl+ku+1`.

The indices 'iJac1=1` and iJac2=10` are used to define the range of matrix elements to output during debugging.

build/source/dshare/popMetadat.f90
martynpclark: The `unknown` dimension can be of many different sizes; too many to list here.

build/source/engine/allocspace.f90
martynpclark: I am a beginner in the use of unlimited polymorphic variables -- I understand that we do need this `select type` construct: Specifically, since `dataStruct` is polymorphic, we need case-appropriate statements. True, the calls to `allocLocal` are identical, though all are needed in this context. Others can test alternative constructs, though I'll stick with this one for now.

build/source/engine/allocspace.f90
martynpclark: See above -- I think that the `select` construct is actually needed here.

build/source/engine/computFlux.f90
martynpclark: Fixed.

build/source/engine/computFlux.f90
martynpclark: Fixed.

build/source/engine/computFlux.f90
martynpclark: Done.

build/source/engine/computFlux.f90
martynpclark: True. Though it is inefficient to initialize to zero when the method is active. Original code retained.

build/source/engine/computFlux.f90
martynpclark: Done.

build/source/engine/computFlux.f90
martynpclark: Fixed.

build/source/engine/computJacob.f90
martynpclark: Yes, the logic is intentional. The derivative in the flux w.r.t. the layer above does not exist for the top layer. Similarly, the derivative in the flux w.r.t. the layer below does not exist for the bottom layer.

build/source/engine/computResid.f90
martynpclark: Fixed

build/source/engine/matrixOper.f90
martynpclark: Fixed.

bartnijssen:
@greyNearing : Grey - Martyn's PR has been merged, please go ahead and merge the latest develop into yours and issue a PR

PR #148, opened 2016-05-09
Bug fix for compilation on hydro-c1 (ifort)

Musselman:
Sorry, this was meant as a pull request to NCAR:develop

Bug fix to compile correctly on hydro-c1 with ifort compiler. In a number of places, the 'select type' construct appeared as 'selecttype' (without space) and the 'type is' statement appeared as 'typeis' (without spaces). Including the spaces permitted compilation.


anewman89:
I think something went wrong Keith.  There should only be one or two files that needed changing with this bug fix.

anewman89:
This should be against NCAR/develop :)  That's the problem.

PR #149, opened 2016-05-09
Bug fix to compile correctly on hydro-c1 with ifort compiler

Musselman:
The 'select type' construct appeared as 'selecttype' (without space)
The 'type is' statement appeared as 'typeis' (without spaces)


bartnijssen:
@Musselman: I think @greyNearing has included this in #153, which is about to be merged. Leave it open for now. I will check after we merge #153

anewman89:
@Musselman @bartnijssen Did this get resolved with #153?

PR #163, opened 2016-06-14
Feature/resume failed solver

luckymichael:
In this Pull Request, an option of resuming model run when the solver does NOT converge. To enable it, add `-r` or `--resume` in the command line.

This is a temporary fix that applies the solution of the last iteration that does not meet the closure criteria.


greyNearing:
Went over it. Logic and syntax looks ok to me.

PR #170, opened 2016-08-15
Hotfix/compact snow liquid drain

luckymichael:
This Pull Request resolves the liquid water locking up in a snow layer when the ice content is higher than a threshold value. It can lead to the error of high snow density in the `snwDensify` subroutine as the immobile liquid becomes ice later at low temperature. Originally, no liquid drainage would occur even when liquid content is high in a dense snow layer. In this fix, we changed that liquid drainage occurs in the same way it does before the ice content is high and inflow is considered to bypass this snow layer.


bartnijssen:
@luckymichael : I'll accept the PR, but please open an issue that we still need to resolve the derivatives. You can refer to this PR in that issue.

PR #172, opened 2016-08-16
Fixed linux build on travis

bartnijssen:
Expanded the travis continuous integration:
 * "manually build" netcdf-fortran using gfortran-6
 * this requires building szip, hdf5 and netcdf-c with the same compiler
 * add expanded scripts for different stages of the compilation

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -136,7 +136,7 @@ NRPROC = $(patsubst %, $(ENGINE_DIR)/%, $(SUMMA_NRPROC))
 
 # Hook-up modules (set files and directory paths)
 SUMMA_HOOKUP= \
-    summaFileManager.f90
+		summaFileManager.f90
 HOOKUP = $(patsubst %, $(HOOKUP_DIR)/%, $(SUMMA_HOOKUP))
 
 # Data modules
@@ -148,7 +148,7 @@ SUMMA_DATAMS= \
 		get_ixname.f90 \
 		ascii_util.f90 \
 		popMetadat.f90 \
-  	outpt_stat.f90
+		outpt_stat.f90
 DATAMS = $(patsubst %, $(DSHARE_DIR)/%, $(SUMMA_DATAMS))
 
 # utility modules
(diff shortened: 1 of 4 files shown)


bartnijssen:
@jhamman : Could you give this a quick review since you are most familiar with the travis-ci system? I modeled this on how it is done for VIC (but currently just test a single compiler and operating system). The build takes fairly long (about 8 mins) because I have to build a whole bunch of packages "by hand".

build/Makefile
```diff
@@ -122,7 +122,7 @@ ENGINE_DIR = $(F_KORE_DIR)/engine
 
 #  utilities
 SUMMA_NRUTIL= \
-        nrtype.f90 \
+		nrtype.f90 \
```
jhamman: There are a number of whitespace only changes in the makefile.  that is fine but I would suggest standardizing on how you wrap lines.  I like this one but others are less tidy.

ci/summa_install_utils
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+
+set -e
```
jhamman: a warning on this (I know we have this still in VIC).

As was pointed out in this https://github.com/travis-ci/travis-ci/issues/6381#issuecomment-236289178:

> Since you are sourcing a bash script that sets errexit ..., the first nonzero exit (from wget) halts the build immediately and you get no further output.

The nice thing about setting `errexit` is that the build stops immediately, the not nice thing is that it is pretty hard to debug in this fashion.

jhamman:
@bartnijssen - looks good to me.  a couple small comments but nothing major.

ci/summa_install_utils
bartnijssen: Won't fix for now, because the build won't succeed without the wget anyway

build/Makefile
bartnijssen: Fixed the whitespace so it is now consistent.

PR #181, opened 2016-08-20
Feature/clean up

martynpclark:
Minor cleanup of the command-line arguments. I thought that it made sense to give every option a switch, as there was unexpected behavior in the `develop` branch as the master file is given as the last command line argument (and all previous command line arguments without a switch were overwritten [and hence ignored]).

### DIFF
--- a/build/source/driver/multi_driver.f90
+++ b/build/source/driver/multi_driver.f90
@@ -1057,7 +1057,7 @@ call stop_program('finished simulation successfully.')
 contains
 
  ! **************************************************************************************************
- ! private function to obtain the command line arguments
+ ! internal function to obtain the command line arguments
  ! **************************************************************************************************
  subroutine getCommandArguments()
  implicit none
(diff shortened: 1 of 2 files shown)


build/source/driver/multi_driver.f90
```diff
@@ -1156,24 +1159,24 @@ subroutine getCommandArguments()
...
- ! private subroutine to print the correct command line usage of SUMMA
+ ! internal subroutine to print the correct command line usage of SUMMA
  ! ************************************************************************************************** 
  subroutine printCommandHelp()  
  implicit none
  ! command line usage
- print "(//A)",'Usage: summa.exe master_file [-r] [-s file_suffix] [-g startGRU countGRU] [-c checkHRU]'
- print "(A)",  ' summa.exe          summa executable'
- print "(A,/)",' master_file        path/name of master file'
+ print "(//A)",'Usage: summa.exe master_file [-m master_file] [-s file_suffix] [-g startGRU countGRU] [-c checkHRU] [-r resume]'
```
luckymichael: remove the first `master_file`

martynpclark: Fixed. Thanks for catching this.

PR #182, opened 2016-08-22
Feature/colbeck test case

martynpclark:
Revisions to the code to read NetCDF initial conditions and NetCDF input, to (1) simplify the code; (2) generalize the code; and (3) add more error checks.


build/source/netcdf/read_icond.f90
```diff
@@ -185,71 +186,78 @@ subroutine read_icond(iconFile,                      & ! name of initial conditi
...
+      message=trim(message)//"unexpectedVariableType[name='"//trim(prog_meta(iVar)%varName)//"';type='"//trim(get_varTypeName(prog_meta(iVar)%varType))//"']"
+      err=20; return
     end select
 
    end do ! iHRU
   end do ! iGRU
 
   ! deallocate storage vector for next variable
-  deallocate(varData)
+  deallocate(varData, stat=err)
```
anewman89: Does this need an error check after?

martynpclark: Fixed, thanks for catching that.

martynpclark:
This PR for the Colbeck test cases is now complete.

bartnijssen:
@luckymichael : Can you review and run the case to make sure it works for someone other than @martynpclark? If it passes I'll go ahead and merge.

build/source/dshare/outpt_stat.f90
```diff
@@ -65,6 +65,7 @@ subroutine calcStats(stat,dat,meta,iStep,err,message)
   if (meta(iVar)%varType==iLookVarType%outstat) then
 
    ! don't do anything if var is not requested
+   ! WHY is this needed when included above?
    if (meta(iVar)%outFreq<0) cycle
```
greyNearing: No, not needed.

build/source/dshare/popMetadat.f90
```diff
@@ -819,100 +846,105 @@ subroutine popStat(meta,lineWords,nWords,indexFlags,err,message)
...
-  if ((.not.found).and.(nFreq.lt.maxFreq)) then          ! if it doesn't exist and we have room, put it in
+  ! identify index of oFreq witin outFreq (cFreq=0 if oFreq is not in outfreq)
+  if(nFreq>0)then
+   cFreq=findIndex(outFreq(1:nFreq),oFreq)  ! index of oFreq, cFreq=0 if not in oFreq
+  else
+   cFreq=0
+  endif
+
+  ! index not found: define index
+  if(cFreq==0)then
```
greyNearing: Note to Grey: Check to make sure this is not initialized to zero.

build/source/netcdf/read_icond.f90
```diff
@@ -185,71 +186,79 @@ subroutine read_icond(iconFile,                      & ! name of initial conditi
  err = nf90_inq_dimid(ncID,"hru",dimID);               if(err/=nf90_noerr)then; message=trim(message)//'problem finding hru dimension/'//trim(nf90_strerror(err)); return; end if
  err = nf90_inquire_dimension(ncID,dimID,len=fileHRU); if(err/=nf90_noerr)then; message=trim(message)//'problem reading hru dimension/'//trim(nf90_strerror(err)); return; end if
 
- ! is there any snow in initial conditions file?
```
greyNearing: Why did this come out? Will it work if there are no snow layers in the restart file?
When I was testing this, it failed if it tried to look for snow data where there was none.

build/source/netcdf/read_icond.f90
```diff
@@ -185,71 +186,79 @@ subroutine read_icond(iconFile,                      & ! name of initial conditi
...
+  endif
 
   ! get variable dimension IDs
   select case (prog_meta(iVar)%varType)
    case (iLookVarType%scalarv); err = nf90_inq_dimid(ncID,trim(scalDimName)   ,dimID); call netcdf_err(err,message)
-   case (iLookVarType%wlength); err = nf90_inq_dimid(ncID,trim(specDimName)   ,dimID); call netcdf_err(err,message)
    case (iLookVarType%midSoil); err = nf90_inq_dimid(ncID,trim(midSoilDimName),dimID); call netcdf_err(err,message)
    case (iLookVarType%midToto); err = nf90_inq_dimid(ncID,trim(midTotoDimName),dimID); call netcdf_err(err,message)
    case (iLookVarType%ifcToto); err = nf90_inq_dimid(ncID,trim(ifcTotoDimName),dimID); call netcdf_err(err,message)
-   case (iLookVarType%midSnow); err = nf90_inq_dimid(ncID,trim(midSnowDimName),dimID); call netcdf_err(err,message)
```
greyNearing: So, how do we initialize the snowpack from a restart file?

build/source/netcdf/read_icond.f90
martynpclark: There are no variables specific to snow: The state variables in the snow+soil domain (temperature, ice content liquid content) are defined for all layers. The only exception is matric head, which is defined only for soil layers.

build/source/netcdf/read_icond.f90
martynpclark: State variables for snow are also state variables for soil and have the dimension `midToto`, This should work. Note that we return an error if `midSnow` is desired and not found.

martynpclark: I have tested this for the Colbeck1976 test case which initializes the snowpack from a restart file.

build/source/dshare/popMetadat.f90
```diff
@@ -819,100 +846,105 @@ subroutine popStat(meta,lineWords,nWords,indexFlags,err,message)
...
-    end if
-    found = .true.
-    exit
-   end if
-  end do
- end do
+ if(count(meta%statFlag)>0)then
+  message=trim(message)//'model output for variable '//trim(meta%varName)//' already defined (can only be defined once)'
+  err=20; return
+ endif
```
luckymichael: is it necessary to be a fatal error? What if the user wants to write both monthly and daily runoff.

bartnijssen: @martynpclark : Please clarify

martynpclark: @luckymichael This was simply cleaning up code from @greyNearing. We currently have the restriction that there is a single timescale for each variable, with the understanding that the use can always temporally aggregate using `nco` commands.

I'm not sure about the reason for this restriction, likely to simplify the metadata structures. @greyNearing wrote the original code and he could comment.

bartnijssen:
Since the answer to the last question does not affect this PR I will just accept it. Would still be good to know the answer.

PR #185, opened 2016-08-29
Feature/split operations

martynpclark:
Implementation of operator splitting. Breaks the full problem into a set of sub-problems that should be easier to solve. Will ultimately enable numerically sub-optimal solutions for cases where the full problem fails to converge.


build/source/driver/multi_driver.f90
```diff
@@ -1192,33 +1202,34 @@ subroutine handle_err(err,message)
...
-   write(*,'(a,100(f11.5,1x))') 'mLayerVolFracIce   = ', progStruct%gru(iGRU)%hru(iHRU)%var(iLookPROG%mLayerVolFracIce)%dat
-   write(*,'(a,100(f11.5,1x))') 'mLayerVolFracLiq   = ', progStruct%gru(iGRU)%hru(iHRU)%var(iLookPROG%mLayerVolFracLiq)%dat
-   print*, 'mLayerMatricHead   = ', progStruct%gru(iGRU)%hru(iHRU)%var(iLookPROG%mLayerMatricHead)%dat
-   print*, 'column inflow      = ', fluxStruct%gru(iGRU)%hru(iHRU)%var(iLookFLUX%mLayerColumnInflow)%dat
-  endif  ! if HRU is valid
- endif  ! if GRU is valid
+ ! dump variables
+ print*, 'error, variable dump:'
+ if(allocated(timeStruct%var))then
+  if(iGRU<=nGRU)then
```
luckymichael: iGRU > 1 and iHRU > 1 can skip dumping variables when error occurs before data structure allocation (even it is very rare). Otherwise it will raise another error as `****Struct` is not allocated.

build/source/engine/systemSolv.f90
```diff
@@ -339,292 +383,529 @@ real(dp)                        :: fOld,fNew                    ! function value
...
+    case(massSplit); nSubstep=1
+    case default; err=20; message=trim(message)//'unable to identify splitting operation'; return
+   end select
+  endif
+
+  ! define the length of the substep
+  dtSplit = dt/real(nSubstep, kind(dp))
+
+  ! loop through substeps
+  do iSubstep=1,nSubstep
```
luckymichael: Will 'nSubstep' always ` == 1` in future? If there are more then one substep, will it be more accurate that place `iSplit` loop in `iSubstep` loop, so every substep the `stateVars (iSplit == 1)` get updated when solving `stateVars (iSplit == 2)`.

martynpclark: Hi @michaelou — the intent is to have variable-dependent sub-stepping at this point. The dynamic substepping for the model as a whole is handled in `coupled_em.f90`

build/source/engine/systemSolv.f90
```diff
@@ -339,292 +383,529 @@ real(dp)                        :: fOld,fNew                    ! function value
...
+
+   ! -----
+   ! * populate state vectors...
+   ! ---------------------------
+
+   ! initialize the global print flag
+   globalPrintFlag=.false.
+
+   ! initialize state vectors
+   call popStateVec(&
```
luckymichael: maybe wrap stateVec into a structure:

```
type stateVar
 physics = [ix_hyd, ix_nrg]
 spatial = [ix_canair, ix_canopy, ix_snow, ix_soil]
 state = [ix_wat, ix_temp, ix_liq, ix_mat]
 value
 xscal
 ...
end type
```

martynpclark: @luckymichael -- do you mean that we should define a structure to hold all of the named variables?

martynpclark: Can you please elaborate?

luckymichael: yes, a structure to hold all of the named variables seems to be more clear to me. and maybe help bookkeeping as they are stored in one place.

build/source/engine/getVectorz.f90
```diff
@@ -149,68 +194,137 @@ real(dp),intent(out)            :: fScale(:)              ! function scaling vec
...
+ ! NOTE: ixVolFracWat  and ixVolFracLiq can also include states in the soil domain, hence enable primary variable switching
+ if(nSnowSoilHyd>0)then
+  do concurrent (iLayer=1:nLayers,ixSnowSoilHyd(iLayer)/=integerMissing)   ! (loop through non-missing hydrology state variables in the snow+soil domain)
+   ixStateSubset                                  = ixSnowSoilHyd(iLayer)   ! index within the state vector
+   select case( ixHydType(iLayer) )
+    case(iname_watLayer); stateVec(ixStateSubset) = mLayerVolFracWat(iLayer)           ! total water state variable for snow+soil layers
+    case(iname_liqLayer); stateVec(ixStateSubset) = mLayerVolFracLiq(iLayer)           ! liquid water state variable for snow+soil layers
+    case(iname_matLayer); stateVec(ixStateSubset) = mLayerMatricHead(iLayer-nSnow)     ! total water matric potential variable for soil layers
+    case(iname_lmpLayer); stateVec(ixStateSubset) = mLayerMatricHeadLiq(iLayer-nSnow)  ! liquid matric potential state variable for soil layers
+    case default; cycle
```
greyNearing: These `cycle` from concurrent do loops do not work in ifort 14.X.X:
`A CYCLE or EXIT statement must not be used with a non-DO block`

The ifort documentation says that cycle will work with concurrent do loops: https://software.intel.com/en-us/node/526121

There is a new 2008 standard about exiting from case selects using `exit` (http://stackoverflow.com/questions/25676793/break-from-select-case), and I wonder if they have defaulted the behavior of the compiler to check for `cycles` instead of `exits` from case selects. 

Either way, we will probably want to change this to a more universal approach - for example, using an if block around the `stateFlag` assignment.

build/source/engine/getVectorz.f90
```diff
@@ -222,149 +336,156 @@ end subroutine popStateVec
...
+ ! overwrite with the hydrology values from the state vector
+ if(nSnowSoilHyd>0)then
+  do concurrent (iLayer=1:nLayers,ixSnowSoilHyd(iLayer)/=integerMissing)   ! (loop through non-missing hydrology state variables in the snow+soil domain)
+   if(globalPrintFlag .and. iLayer==101) write(*,'(a,1x,10(i4,1x))') 'iLayer, ixHydType(iLayer), ixSnowSoilHyd(iLayer) = ', iLayer, ixHydType(iLayer), ixSnowSoilHyd(iLayer)
+   select case( ixHydType(iLayer) )
+    case(iname_watLayer); mLayerVolFracWatTrial(iLayer)          = stateVec( ixSnowSoilHyd(iLayer) ) ! total water state variable for snow+soil layers
+    case(iname_liqLayer); mLayerVolFracLiqTrial(iLayer)          = stateVec( ixSnowSoilHyd(iLayer) ) ! liquid water state variable for snow+soil layers
+    case(iname_matLayer); mLayerMatricHeadTrial(iLayer-nSnow)    = stateVec( ixSnowSoilHyd(iLayer) ) ! total water matric potential variable for soil layers
+    case(iname_lmpLayer); mLayerMatricHeadLiqTrial(iLayer-nSnow) = stateVec( ixSnowSoilHyd(iLayer) ) ! liquid matric potential state variable for soil layers
+    case default; cycle
```
greyNearing: See above.

build/source/engine/getVectorz.f90
```diff
@@ -149,68 +194,137 @@ real(dp),intent(out)            :: fScale(:)              ! function scaling vec
...
+  stateFlag( ixVegNrg(iState) ) = .true.                 ! flag to denote that the state is populated
+ end do
+
+ ! build the state vector for the water in the vegetation canopy
+ ! NOTE: currently vector length=1, and use "do concurrent" to generalize to a multi-layer canopy
+ do concurrent (iState=1:size(ixVegHyd),ixVegHyd(iState)/=integerMissing)
+  select case(ixStateType_subset( ixVegHyd(iState) ))
+   case(iname_watCanopy); stateVec( ixVegHyd(iState) )  = scalarCanopyWat        ! transfer total canopy water to the state vector
+   case(iname_liqCanopy); stateVec( ixVegHyd(iState) )  = scalarCanopyLiq        ! transfer liquid canopy water to the state vector
+   case default; cycle
```
greyNearing: See next comment.

build/source/engine/updateVars.f90
```diff
@@ -0,0 +1,631 @@
...
+    case(iname_veg, iname_snow); Tcrit = Tfreeze
+    case(iname_soil);            Tcrit = crit_soilT( mLayerMatricHeadTrial(ixControlIndex) )
+    case default; err=20; message=trim(message)//'expect case to be iname_veg, iname_snow, iname_soil'; return
+   end select
+  
+   ! (compute the derivative in liquid water content w.r.t. temperature)
+   select case(ixDomainType)
+    case(iname_veg);  dTheta_dTkCanopy         = merge(dFracLiq_dTk(xTemp,snowfrz_scale)*scalarCanopyWat/(iden_water*canopyDepth),  0._dp, xTemp<Tcrit)
+    case(iname_snow); mLayerdTheta_dTk(iLayer) = merge(dFracLiq_dTk(xTemp,snowfrz_scale)*mLayerVolFracWatTrial(iLayer),             0._dp, xTemp<Tcrit)
+    case(iname_soil); mLayerdTheta_dTk(iLayer) = merge(dTheta_dTk(xTemp,theta_res,theta_sat,vGn_alpha,vGn_n,vGn_m),                 0._dp, xTemp<Tcrit) ! assume no volume expansion
```
greyNearing: As far as I can tell, `function dTheta_dTk` was never called in previous versions - although this function does already exist in `soils_util.f90`. In that function, if `vGn_alpha < 0` and temperatures are above freezing, then the function takes a fractional power of a negative number and tries to store the result as a real(dp). 

I am not sure why it was allowing you to do this with gfortran ...

bartnijssen:
@martynpclark and @luckymichael : Do we still need this PR or has it been superseded by PR 195 (which includes most of the commits here, but not all).

luckymichael:
@bartnijssen I believed this has been superseded by PR 195.

bartnijssen:
@martynpclark : I am planning to merge this one before #195 to make sure that we have the split operations PR separate from the work we did for the Columbia

PR #192, opened 2016-10-14
ooperator splitting solution with Explicit Euler as last resort

martynpclark:
This code is a major refactor to enable operator splitting (i.e., solving a series of smaller sub-problems as an alternative to the fully coupled solution). The code also implements the explicit Euler solution for cases where the smallest sub-problem fails to converge.


luckymichael:
replaced by PR 195

bartnijssen:
Closed since replaced by PR 195

PR #193, opened 2016-10-18
fix the bug in read_force()

luckymichael:
fix the bug that it does not read time unit after the first forcing file in read_force


luckymichael:
merged in PR195

PR #194, opened 2016-10-20
Hotfix/rename spinup ouput

luckymichael:
1. rename the spinup output file name based on start time to prevent the file from being overwritten when multiple restart runs are performed
2. change the starting time of a new water year time from Oct 1 1:00 to Oct 1 0:00

### DIFF
--- a/build/source/driver/multi_driver.f90
+++ b/build/source/driver/multi_driver.f90
@@ -662,9 +662,12 @@ select case (iRunMode)
  case(iRunModeHRU)
   write(output_fileSuffix((len_trim(output_fileSuffix)+1):len(output_fileSuffix)),"('_H',i0)") checkHRU
 end select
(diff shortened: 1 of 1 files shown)


luckymichael:
@martynpclark : Need to clarify time stamps (period-ending or period-starting) before we proceed with this PR.

luckymichael:
@martynpclark and @bartnijssen : I realized that whether it is period-ending or period-starting also depends on how the time is defined in the forcing files. If you use the ending time for each time step in the forcing file, the water year will start from Oct 1 1:00 otherwise Oct 1 0:00. Therefore, it depends on whether you prefer to use 0:00 or 24:00 to split the days. For me, 0:00 makes more sense because it is compatible with the tools I often used to handle the output files (python, R, etc).

PR #196, opened 2016-10-23
associate two soil water depth variables to store their values

luckymichael:


luckymichael:
@bartnijssen : go ahead and merge

PR #202, opened 2017-02-23
Develop

martynpclark:
Ensures that the explicit Euler update is less than a specified threshold.


martynpclark:
I reset error code for cases of recoverable errors (to avoid the bug where we had a graceful exit for no good reason).

PR #203, opened 2017-03-08
define max line width for assumed-shape character arrays

ougx:
avoid assumed-length character variables when using Intel® Math Kernel Library LAPACK

### DIFF
--- a/build/source/dshare/ascii_util.f90
+++ b/build/source/dshare/ascii_util.f90
@@ -21,6 +21,7 @@
 module ascii_util_module
 USE nrtype
 implicit none
+integer(i4b),parameter,public               :: linewidth=4096          ! max character number for one line
 private
 public::file_open
 public::split_line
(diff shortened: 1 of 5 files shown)


martynpclark:
This looks good to me. However I do not see the relationship to Intel® Math Kernel Library LAPACK. Can you lease clarify why this was done? It is not clear to me what problem this solves.

bartnijssen:
@luckymichael as requested by @martynpclark, please provide some context

luckymichael:
This is actually a bug with `ifort 17` that it cannot correctly detect the length of the outgoing character array when it is an allocatable assumed-length character array. Instead the length is determined based on the content assigned to the variable. And therefore the assignment will be done compactly (continuously) in memory without `space`s between elements. The content will fill the whole length of the first elements of the returning array sequentially.

I realized that it's not related to Intel® Math Kernel Library LAPACK but I happened to use `ifort 17` when using MKL. To use MKL, use these options in the Makefile:
```
LAPK_PATH = ${MKLROOT}
LIBLAPACK = -L$(LAPK_PATH)/lib/intel64 -lmkl_rt
```
MKL should be shipped with the Intel Compiler and it is claimed that it is more efficient than the standard LAPACK package when using `ifort`.

luckymichael:
close because it is a bug with a specific `ifort` version which I assumed it will be fixed by Intel.

arbennett:
This still is a problem.  Could we re-open?  Issued a pull request that fixes this #219

bartnijssen:
@arbennett: Did you really mean that I should re-open the PR? May make more sense to open an issue that refers to both (which is why I like to have an issue separate from the PR).

arbennett:
Oops, I thought this was an issue, not a PR.  I guess if you search for something under the issues tab it also shows pull requests...

bartnijssen:
Yes it does - so if I now close it, will it "cancel" the PR (as in undo it?). This also underscores why I like to have a separate issue for each PR. We should be able to reopen an issue without messing with the PR.

bartnijssen:
Closed again - superceeded by PR #219

PR #204, opened 2017-03-08
Docker

schymans:
We (@jirikuncar and I) used a minimalistic ubuntu base image and made a few changes to the makefile so that it can run in the docker image without any modifications. 
Summa actually compiles now using the native netcdf libraries, so the ci/summa_install_utils is not needed for that. We made changes there while trying to use it, but found a simpler solution. 
To create the docker image, just run `docker build .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM phusion/baseimage:0.9.19
+
+USER root
+
+RUN apt-get update && \
+    add-apt-repository ppa:ubuntu-toolchain-r/test && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    build-essential \
+    libnetcdff-dev \
+    netcdf-bin \
+    liblapack-dev \
+    gfortran-6 \
+    gcc && apt-get clean
+
+ENV FC gfortran-6
+ENV F_MASTER /code
+ENV NCDF_PATH /usr
+ENV LD_LIBRARY_PATH ${NCDF_PATH}/lib
+
+WORKDIR /code
+ADD . /code
+
+RUN make -C build/ -f Makefile
--- a/build/Makefile
+++ b/build/Makefile
@@ -13,11 +13,11 @@
 
 # Define core directory below which everything resides. This is the
 # parent directory of the 'build' directory
-F_MASTER =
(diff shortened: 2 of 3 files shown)


bartnijssen:
The build breaks on travis because you hardcoded the compiler and flags in the Makefile. The `.travis.yml` file configures does specifically. It's good to see your Dockerfile though. I will leave the PR open for now, but will probably replace it by another one once I have a build that works for both travis and Docker. 

@schymans : One question - is there a specific reason for using `phusion/baseimage:0.9.19`? I was planning to use `ubuntu:zesty`, but don't want to break anything for you.

schymans:
Thanks for the clarification. I thought the travis failure was due to our original modification of CI/summa_install_utils, which we reverted again. We didn't actually hard-code the variables in the makefile but removed the un-initiated lines and provided them as ENV in the Dockerfile:
`ENV FC gfortran-6
ENV F_MASTER /code
ENV NCDF_PATH /usr
ENV LD_LIBRARY_PATH ${NCDF_PATH}/lib`
This was to us a more natural way than to have to edit the makefile for each platform. I think if you add ENV FC and ENV F_MASTER to the .travis file, it should work.
I used phusion/baseimage to have a minimal ubuntu and not use all the unnecessary bits that come with ubuntu:zesty. Read here: [https://github.com/phusion/baseimage-docker](https://github.com/phusion/baseimage-docker)

bartnijssen:
That works on Docker, but the travis build works slightly different. I have a working Docker build (for which your file was helpful), but right now I have a problem pulling that image from `hub.docker.com` (not sure what is going on - this is new to me as well). As soon as I have that resolved I should be able to share it with you.

jirikuncar:
@bartnijssen I have fixed the problem with Travis. It should be easy to setup automatic push to Docker (see e.g. https://sebest.github.io/post/using-travis-ci-to-build-docker-images/) for each commit and release. Please ping me if you have any problems.

bartnijssen:
Thanks @jirikuncar : I'll accept the PR as-is and will then make some small changes. I am planning to change the Makefile a bit, but this definitely helped me understand the Dockerfile much better.

Issue #214, opened 2017-08-29
Restart file check for `_fillValue` is too strict


bartnijssen:
The check in https://github.com/NCAR/summa/blob/develop/build/source/netcdf/read_icond.f90#L229-L233
is too strict. We are checking that **no** values in `varData` equal `nf90_fill_double`, but there are cases where we should expect `nf90_fill_double`.

In particular, if there are multiple hrus in the file, then for some variables `varData` is sized to accommodate the maximum number of soil+snow layers across all hrus. For hrus that have fewer total layers (for example because they have fewer snow layers), `varData` will be padded with `nf90_fill_double`.

If we want to maintain the check, then we should place it within the `select ... case` statements in https://github.com/NCAR/summa/blob/develop/build/source/netcdf/read_icond.f90#L245-L254, because in that case `nSoil` and `nSnow` are specific to each hru.

As part of this we should also remove the comment:
https://github.com/NCAR/summa/blob/develop/build/source/driver/multi_driver.f90#L535
since I think that is taken care of properly in `read_icond.f90`.

martynpclark:
@bartnijssen -- is this issue now resolved?

PR #215, opened 2017-08-29
Fix/issue214

bartnijssen:
Move check for `nf90_fill_double` as explained in issue #214


bartnijssen:
There are a lot of whitespace changes in this, which make this looking much more involved than it is. Ignore the whitespace changes. See also #216

martynpclark:
Can we have just one PR that removes white spaces (throughout the entire code) and one PR that actually makes code changes?

bartnijssen:
Re-issued as a new PR without the whitespace changes. See #250

PR #219, opened 2017-08-31
Bugfix/icc17 linewidth

### LINKED ISSUE #222: Using `ifort 17` causes crashes when reading output configuration
As per @ougx in #203: 

>  This is actually a bug with ifort 17 that it cannot correctly detect the length of the outgoing character  array when it is an allocatable assumed-length character array. Instead the length is determined based on the content assigned to the variable. And therefore the assignment will be done compactly (continuously) in memory without spaces between elements. The content will fill the whole length of the first elements of the returning array sequentially.

This can be seen when reading the output configuration file, which has variable length.  The current implementation uses a linked list to read in all of the contents and then is supposed to transfer the data from that list into an allocatable array.  Instead of loading each node into a slot of the array it currently loads everything into the first one, which causes crashes after reading.

arbennett:
Fixes #222
Redux of #203

For clarity: I reimplemented this because the merge conflicts would have been harder to resolve than just fixing an updated version of the code.

### DIFF
--- a/build/source/dshare/ascii_util.f90
+++ b/build/source/dshare/ascii_util.f90
@@ -21,6 +21,7 @@
 module ascii_util_module
 USE nrtype
 implicit none
+integer(i4b),parameter,public :: linewidth=4096  ! max character number for one line
 private
 public::file_open
 public::split_line
@@ -71,10 +72,10 @@ contains
  ! do not know how many "words", so use linked lists
  implicit none
  ! declare dummy arguments
- character(*),intent(in)              :: inline     ! line of characters
- character(*),intent(out),allocatable :: words(:) ! vector of "words"
- integer(i4b),intent(out)             :: err      ! error code
- character(*),intent(out)             :: message  ! error message
+ character(*),intent(in)                          :: inline     ! line of characters
+ character(len=*),intent(out),allocatable         :: words(:)   ! vector of "words"
+ integer(i4b),intent(out)                         :: err        ! error code
+ character(*),intent(out)                         :: message    ! error message
  ! declare local variables
  integer(i4b),parameter  :: cLen=8192
  character(len=cLen)     :: temp                  ! temporary line of characters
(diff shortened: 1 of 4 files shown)


bartnijssen:
@arbennett: But this duplicates a lot of what is already in PR 203 - it's unclear to me what you changed here.

bartnijssen:
Can you just do this as a brief PR that is stand-alone but references #203?

build/source/dshare/ascii_util.f90
```diff
@@ -21,6 +21,7 @@
 module ascii_util_module
 USE nrtype
 implicit none
+integer(i4b),parameter,public :: linewidth=4096  ! max character number for one line
```
bartnijssen: Should this be the same length as in l.80 : `8192`?

build/source/dshare/popMetadat.f90
```diff
@@ -749,6 +750,7 @@ subroutine read_output_file(err,message)
  do vLine = 1,size(charLines)
 
   ! parse the current line
+  print*, trim(charLines(vLine))
```
bartnijssen: Remove debug `print` statements

build/source/dshare/ascii_util.f90
```diff
@@ -134,10 +135,10 @@ subroutine get_vlines(unt,vlines,err,message)
  ! do not know how many valid lines, so use linked lists
  implicit none
  ! declare dummy arguments
- integer(i4b),intent(in)              :: unt         ! file unit
- character(*),intent(out),allocatable :: vlines(:)   ! vector of character strings
- integer(i4b),intent(out)             :: err         ! error code
- character(*),intent(out)             :: message     ! error message
+ integer(i4b),intent(in)                          :: unt         ! file unit
+ character(len=linewidth),intent(out),allocatable :: vlines(:)   ! vector of character strings
```
bartnijssen: So - this is the heart of the matter, correct?

bartnijssen:
Can you make sure this still compiles and operates correctly with `gfortran`? We know it compiles from travis, but we need to build in some checks

build/source/dshare/ascii_util.f90
arbennett: In some sense it's arbitrary.  It might be nice to have lengths like this be consistent, but I hadn't thought about it when implementing this.

build/source/dshare/ascii_util.f90
arbennett: Correct.

PR #220, opened 2017-08-31
Adjust canopy windspeed and resistance contribution with snow depth

arbennett:
The `aeroResist` subroutine currently does not take the snow depth into account with calculating the wind speeds at the top and bottom of the canopy, as well as when calculating the ground resistance under neutral conditions.  This pull request adjusts those calculations and handles cases where the snow covers the canopy bottom as well as when snow covers the canopy completely.

### DIFF
--- a/build/source/engine/vegNrgFlux.f90
+++ b/build/source/engine/vegNrgFlux.f90
@@ -2078,6 +2078,8 @@ contains
  real(dp)                      :: windConvFactorBottom          ! factor to convert wind speed at top of canopy to wind speed at bottom of canopy (-)
  real(dp)                      :: referenceHeight               ! reference height used to compute above-ground windspeed (m)
  real(dp)                      :: heightAboveGround             ! height above the snow surface (m)
+ real(dp)                      :: heightCanopyTopAboveSnow      ! height at the top of the vegetation canopy relative to snowpack (m)
+ real(dp)                      :: heightCanopyBottomAboveSnow   ! height at the bottom of the vegetation canopy relative to snowpack (m)
  ! local variables: derivatives
  real(dp)                      :: dFV_dT                        ! derivative in friction velocity w.r.t. canopy air temperature
  real(dp)                      :: dED_dT                        ! derivative in eddy diffusivity at the top of the canopy w.r.t. canopy air temperature
@@ -2108,6 +2110,8 @@ contains
  if(computeVegFlux) then ! (if vegetation is exposed)
 
   ! ***** identify zero plane displacement, roughness length, and surface temperature for the canopy (m)
+  heightCanopyTopAboveSnow = heightCanopyTop - snowDepth
+  heightCanopyBottomAboveSnow = max(heightCanopyBottom - snowDepth, 0.0_dp)
   select case(ixVegTraits)
 
    ! Raupach (BLM 1994) "Simplified expressions..."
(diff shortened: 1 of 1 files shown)


bartnijssen:
@arbennett - can you open an issue for this in which you provide some background for the problem and some of the figures you made. You can then refer to that issue in the PR. I typically like to have one or more issues per PR, since the PR is really about the code. So workflow: open issue - then fix and open PR that references the issue. It's easier to keep track of things that way.

build/Makefile
```diff
@@ -16,7 +16,7 @@
 ##F_MASTER =
 
 # Define the Fortran Compiler. If you are using gfortran, then this needs
-# to be version 4.8 or higher. This variable is simply used to select the right
+# to be version 6 or higher. This variable is simply used to select the right
```
bartnijssen: Why does this keep showing up? This has nothing to do with this PR?

arbennett: I merged master into develop, the two branches are not synced up.

arbennett:
This fixes #223

arbennett:
I have created a notebook showing the output from the code included in this pull request: https://gist.github.com/arbennett/8914bc10b3c0fde605cd4234c9b83eb8

martynpclark:
@arbennett -- I see that the canopy resistance is constant once the canopy is completely covered by snow. Would it be clearer if canopy resistance is set to missing in this case?

build/source/engine/vegNrgFlux.f90
```diff
@@ -2104,60 +2106,54 @@ subroutine aeroResist(&
...
  ! -----------------------------------------------------------------------------------------------------------------------------------------
  ! * compute vegetation poperties (could be done at the same time as phenology.. does not have to be in the flux routine!)
-
  if(computeVegFlux) then ! (if vegetation is exposed)
 
   ! ***** identify zero plane displacement, roughness length, and surface temperature for the canopy (m)
+  ! First, calculate offset from snow depth - use these to
+  ! scale wind profiles and resistances
+  heightCanopyTopAboveSnow = heightCanopyTop - snowDepth
+  heightCanopyBottomAboveSnow = max(heightCanopyBottom - snowDepth, 0.0_dp)
```
martynpclark: I am not sure if this is correct, especially the truncation at zero.

build/source/engine/vegNrgFlux.f90
```diff
@@ -2191,66 +2187,49 @@ subroutine aeroResist(&
...
-  windReductionFactor = windReductionParam * exposedVAI**twoThirds * (heightCanopyTop - heightCanopyBottom)**oneThird / leafDimension**oneThird
-  !windReductionFactor = 3._dp
+  windReductionFactor = windReductionParam * exposedVAI**twoThirds * (heightCanopyTopAboveSnow - heightCanopyBottomAboveSnow)**oneThird / leafDimension**oneThird
 
   ! compute windspeed at the bottom of the canopy (m s-1)
-  !referenceHeight      = max(heightCanopyBottom, min(0.5_dp, heightCanopyTop))
-  referenceHeight      = max(heightCanopyBottom, snowDepth+z0Ground)
-  windConvFactorBottom = exp(-windReductionFactor*(1._dp - referenceHeight/heightCanopyTop))
+  referenceHeight      = max(heightCanopyBottomAboveSnow, z0Ground)
+  windConvFactorBottom = exp(-windReductionFactor*(1._dp - (referenceHeight/heightCanopyTopAboveSnow)))
```
martynpclark: There calculations are only used when the snow is below the bottom of the canopy. Could you move these calculations to the `case(logBelowCanopy)` below?

arbennett: There is a call to `aStability` which uses `windspdCanopyBottom`, so we actually can't move these inside the `select` statement.

arbennett:
@martynpclark I think it probably would be cleaner to set canopy resistance to `missingValue` when covered, but I'm not sure how to implement it.

PR #224, opened 2017-08-31
Merge master into develop

bartnijssen:
Keep develop in sync with master


bartnijssen:
OK - pushed directly - not sure that is exactly what I intended, but the end result looks good.

PR #228, opened 2017-09-01
Use `class is` on derived type select statements.

### LINKED ISSUE #227: Cannot determine derived type in select type statement with some gfortran builds
With certain versions/builds of `gfortran` I run into this error:
```
FATAL ERROR: coupled_em/opSplittin/allocLocal/unable to identify derived data type for the variable dimension
```
which comes from this chunk of code:
```
 select type(dataStruct)
  type is (var_i);       if(allocated(dataStruct%var))then; check=.true.; else; allocate(dataStruct%var(nVars),stat=err); end if; return
  type is (var_d);       if(allocated(dataStruct%var))then; check=.true.; else; allocate(dataStruct%var(nVars),stat=err); end if; return
  type is (var_ilength); if(allocated(dataStruct%var))then; check=.true.; else; allocate(dataStruct%var(nVars),stat=err); end if
  type is (var_dlength); if(allocated(dataStruct%var))then; check=.true.; else; allocate(dataStruct%var(nVars),stat=err); end if
  class default; err=20; message=trim(message)//'unable to identify derived data type for the variable dimension'; return
 end select
```

The variable that triggers this is from `opSplittin.f90` at line 377:
```
call allocLocal(prog_meta(:),prog_temp,nSnow,nSoil,err,cmessage)
```
Where `prog_temp` is declared to be `type(var_dlength)`.  It seems from [this specification](https://www.ibm.com/support/knowledgecenter/SSGH4D_15.1.2/com.ibm.xlf151.aix.doc/language_ref/typecon.html) that this may occur because we are declaring the `dataStruct` as `class(*)`, and the runtime can't infer the correct type. Switching these statements to `class is` seems to alleviate the issue.

This issue occurred with the following versions of `gfortran`:
```
Using built-in specs.                      
COLLECT_GCC=gfortran                       
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/7.1.1/lto-wrapper                
Target: x86_64-pc-linux-gnu                
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --disable-multilib --disable-werror --enable-checking=release --enable-default-pie --enable-default-ssp
Thread model: posix                        
gcc version 7.1.1 20170630 (GCC)
```

```
Using built-in specs.
COLLECT_GCC=gfortran-6
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/6.4.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: ../configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++,java --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --with-linker-hash-style=gnu --enable-gnu-indirect-function --disable-multilib --disable-werror --enable-checking=release --enable-java-awt=gtk --with-java-home= --enable-libgcj-multifile --enable-version-specific-runtime-libs --program-suffix=-6 --build=x86_64-pc-linux-gnu
Thread model: posix
gcc version 6.4.0 (GCC)
```

arbennett:
This fixes #227. Tested on `gfortran` and `ifort`.

### DIFF
--- a/build/source/dshare/outpt_stat.f90
+++ b/build/source/dshare/outpt_stat.f90
@@ -68,8 +68,8 @@ contains
 
    select type (dat)
     type is (real(dp)); tdata = dat(pVar)
-    type is (dlength) ; tdata = dat(pVar)%dat(1)
-    type is (ilength) ; tdata = real(dat(pVar)%dat(1), kind(dp))
+    class is (dlength) ; tdata = dat(pVar)%dat(1)
+    class is (ilength) ; tdata = real(dat(pVar)%dat(1), kind(dp))
     class default;err=20;message=trim(message)//'dat type not found';return
    end select
 
@@ -124,8 +124,8 @@ contains
 
  ! pack back into struc
  select type (stat)
-  type is (ilength); tstat = real(stat%dat)
-  type is (dlength); tstat = stat%dat
+  class is (ilength); tstat = real(stat%dat)
+  class is (dlength); tstat = stat%dat
   class default;err=20;message=trim(message)//'stat type not found';return
  end select
 
@@ -200,8 +200,8 @@ contains
 
  ! pack back into struc
  select type (stat)
-  type is (ilength); stat%dat = int(tstat)
-  type is (dlength); stat%dat = tstat
+  class is (ilength); stat%dat = int(tstat)
+  class is (dlength); stat%dat = tstat
   class default;err=20;message=trim(message)//'stat type not found';return
  end select
 
--- a/build/source/engine/allocspace.f90
+++ b/build/source/engine/allocspace.f90
@@ -94,15 +94,15 @@ contains
  ! * allocate GRU dimension
  select type(dataStruct)
   ! gru dimension only
-  type is (gru_int);           if(allocated(dataStruct%gru))then; check=.true.; else; allocate(dataStruct%gru(nGRU),stat=err); end if
-  type is (gru_intVec);        if(allocated(dataStruct%gru))then; check=.true.; else; allocate(dataStruct%gru(nGRU),stat=err); end if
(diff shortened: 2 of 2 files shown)


bartnijssen:
@arbennett: Are you sure that all the `type is` need to be replaced by `class is` or did you simply replace them all.

arbennett:
I did not do a full write-compile-run-loop for each of the replacements, but I did start with just the one that was troublesome, got the same error a bit further in when it tried to do the same with one of the other derived types within `allocLocal`, so I replaced all of those statements and re-ran.  This gave an error in `output_stat.f90`. I basically went through the same loop on this file then.

bartnijssen:
I ran all test cases with this PR and compared the resulting files with output from the version before this PR compiled with `gfortran6` on OS X (not sure why that worked and the one for @arbennett did not). Output (metadata and data) was identical for all files.

Closes #227

Issue #256, opened 2017-09-26
Canopy Depth set to Zero


msaharia:
For certain GRUs, the canopy depth is being set to Zero. `PR` incoming on this issue. 

```
computeVegFlux, canopyDepth =  F  0.000000000000000E+000
forrtl: error (73): floating divide by zero
Image              PC                Routine            Line        Source
summa.exe          0000000000AB3A01  Unknown               Unknown  Unknown
summa.exe          0000000000AB2157  Unknown               Unknown  Unknown
libnetcdff.so.5    00007F5CFDEADB32  Unknown               Unknown  Unknown
libnetcdff.so.5    00007F5CFDEAD986  Unknown               Unknown  Unknown
libnetcdff.so.5    00007F5CFDE94B2C  Unknown               Unknown  Unknown
libnetcdff.so.5    00007F5CFDE993BE  Unknown               Unknown  Unknown
libpthread.so.0    00007F5CFD0CC890  Unknown               Unknown  Unknown
summa.exe          00000000009A9376  tempadjust_module         120  tempAdjust.f90
summa.exe          000000000051E188  coupled_em_module         458  coupled_em.f90
summa.exe          000000000078627C  MAIN__                    902  multi_driver.f90
summa.exe          00000000004058FE  Unknown               Unknown  Unknown
libc.so.6          00007F5CFCD33B45  Unknown               Unknown  Unknown
summa.exe          00000000004057B9  Unknown               Unknown  Unknown
Aborted
```

PR #257, opened 2017-09-26
Fix for Issue #256

msaharia:


build/source/engine/coupled_em.f90
```diff
@@ -453,16 +453,18 @@ lightSnow            ! maximum interception capacity an inverse function of new
...
+  call tempAdjust(&
+                  ! input: derived parameters
+                  canopyDepth,                 & ! intent(in): canopy depth (m)
+                  ! input/output: data structures
+                  mpar_data,                   & ! intent(in):    model parameters
+                  prog_data,                   & ! intent(inout): model prognostic variables for a local HRU
+                  diag_data,                   & ! intent(out):   model diagnostic variables for a local HRU
+                  ! output: error control
+                  err,cmessage)                  ! intent(out): error control
+  if(err/=0)then; err=20; message=trim(message)//trim(cmessage); return; end if
```
bartnijssen: Indent the `if(err/=0)` line

bartnijssen: Do we need to skip more than just this routine? E.g. should `canopySnow` also be skipped?

martynpclark:
Other routines pass the `computeVegFlux` flag, and conduct separate calculations for the case of no vegetation in the routines themselves.

It would be possible to do this here also; however, in my opinion it is clearer to have the check outside the subroutine, since the temperature of the canopy is only adjusted if the canopy exists or is exposed.

PR #262, opened 2017-10-13
TESTING NEEDED: Feature/model write

martynpclark:
Modify write statements so that all variables are time unlimited.

NOTE: This compiles and runs, and needs testing.


martynpclark:
PR superseded by #268

PR #271, opened 2017-10-24
Bug fix/deep snow

martynpclark:
This PR supersedes PR #268. It includes a bug fix for the case where snow exceeds the measurement height.

PR #272, opened 2017-10-25
Feature/add HRU id

martynpclark:
Changes name of `hruIndex` to `hruId` to clarify that it is the unique identifier rather than the index in the vector.

This PR supersedes PR #271


arbennett:
This is redundant and was superseded by #275

PR #273, opened 2017-10-25
Feature/deep snow

martynpclark:
consider case where snow exceeds measurement height


martynpclark:
inadvertently made pr to master

PR #274, opened 2017-10-25
Feature/deep snow

martynpclark:
ensure measurement height is higher than the snow depth


arbennett:
This was solved in #275 and #276

PR #285, opened 2017-11-22
Restart file to have same dimensions as Output file

msaharia:
1. Tested for a 3 Layer model 
2. Retro run followed by a forecast run using the concatenated restart file file from retro run

### DIFF
--- a/build/source/netcdf/modelwrite.f90
+++ b/build/source/netcdf/modelwrite.f90
@@ -262,7 +262,7 @@ contains
 
       end do  ! HRU loop
      end do  ! GRU loop
- 
+
      ! get the maximum length of each data vector
      select case (meta(iVar)%varType)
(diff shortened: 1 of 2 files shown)


build/source/netcdf/modelwrite.f90
```diff
@@ -274,7 +274,7 @@ subroutine writeData(modelTimestep,outputTimestep,nHRUrun,maxLayers,meta,stat,da
       case(iLookVarType%ifcSoil); maxLength = nSoil+1
       case default; cycle
      end select ! vartype
-
+
```
martynpclark: Please set editor preferences to remove trailing white space

msaharia: @martynpclark Set. Also, removed from these files.

PR #287, opened 2017-11-26
Bug fix/duplicate meas height

martynpclark:
Fixed temporary change where measurement height was always taken as above the top of the canopy.

Tested for the 100-year run over the Columbia.


bartnijssen:
@arbennett: Can you quickly check this? It's nice and small (thanks @martynpclark)

PR #288, opened 2017-11-26
Feature/add version info and remove duplicate measurement height

martynpclark:
Included version information in model output file.


bartnijssen:
@martynpclark : Would be easier if you had branched from develop rather than from a branch with other changes (like to avoid compounded PRs). Leave as is for now and I'll merge it after #287

PR #291, opened 2017-11-27
Feature/track numerix

martynpclark:
Keeps track of the number of times different splitting operations are used in each time step (as well as the total number of flux calls per step)


bartnijssen:
Most changes just align text

PR #292, opened 2017-11-27
Bug fix/saturated flow

martynpclark:
This PR should be merged after #290 

There are two changes here:
(1) Included tolerance in initial conditions check to account for finite precision;
(2) Reduced the minimum time step for the different numerix options (only use the scalar solution as a last resort).

### DIFF
--- a/build/source/engine/check_icond.f90
+++ b/build/source/engine/check_icond.f90
@@ -90,7 +90,7 @@ contains
  integer(i4b)                      :: nLayers            ! total number of layers
  real(dp)                          :: kappa              ! constant in the freezing curve function (m K-1)
  integer(i4b)                      :: nSnow              ! number of snow layers
-
+ real(dp),parameter                :: xTol=1.e-10_dp     ! small tolerance to address precision issues
  ! --------------------------------------------------------------------------------------------------------
 
  ! Start procedure here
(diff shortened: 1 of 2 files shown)


build/source/engine/check_icond.f90
```diff
@@ -196,14 +196,14 @@ subroutine check_icond(nGRU,                          & ! number of GRUs and HRU
      ! ***** soil
      case(iname_soil)
       ! (check liquid water)
-      if(mLayerVolFracLiq(iLayer) < theta_res(iSoil) )then; write(message,'(a,1x,i0)') trim(message)//'cannot initialize the model with volumetric fraction of liquid water < theta_res: layer = ',iLayer; err=20; return; end if
-      if(mLayerVolFracLiq(iLayer) > theta_sat(iSoil) )then; write(message,'(a,1x,i0)') trim(message)//'cannot initialize the model with volumetric fraction of liquid water > theta_sat: layer = ',iLayer; err=20; return; end if
+      if(mLayerVolFracLiq(iLayer) < theta_res(iSoil)-xTol)then; write(message,'(a,1x,i0)') trim(message)//'cannot initialize the model with volumetric fraction of liquid water < theta_res: layer = ',iLayer; err=20; return; end if
+      if(mLayerVolFracLiq(iLayer) > theta_sat(iSoil)+xTol)then; write(message,'(a,1x,i0)') trim(message)//'cannot initialize the model with volumetric fraction of liquid water > theta_sat: layer = ',iLayer; err=20; return; end if
```
bartnijssen: The other option would be to write a function that takes two arguments and compares that the absolute difference is less than `xTol`. Let's leave you have it here for now.

build/source/engine/opSplittin.f90
```diff
@@ -251,8 +251,8 @@ subroutine opSplittin(&
  ! ------------------------------------------------------------------------------------------------------
  ! minimum timestep
  real(dp),parameter              :: dtmin_coupled=1800._dp         ! minimum time step for the fully coupled solution (seconds)
- real(dp),parameter              :: dtmin_split=360._dp            ! minimum time step for the fully split solution (seconds)
- real(dp),parameter              :: dtmin_scalar=60._dp            ! minimum time step for the scalar solution (seconds)
+ real(dp),parameter              :: dtmin_split=60._dp             ! minimum time step for the fully split solution (seconds)
+ real(dp),parameter              :: dtmin_scalar=10._dp            ! minimum time step for the scalar solution (seconds)
```
bartnijssen: Should we pull these out into some `constants` or `model setup` section? No change required at this time.

PR #294, opened 2017-12-01
Bug fix/gfortran7

martynpclark:
Change "type is" to "class is"


bartnijssen:
@martynpclark: Next time, can you do a PR for a big fix like this from develop rather than from your branch that has both the wind profile and the bugfix changes in it? I.e. when you branch to make a bug fix, don't branch from your current branch but start with one that does not have unmerged changes. It makes life easier for everyone. That is, this PR should only have commits 
f257de9 and d2aa8f3. Leave as is for now. We'll first merge the wind profile one and then this one should be clean.

PR #300, opened 2018-01-06
Feature/nldas adjustments

martynpclark:
Minor adjustments to run over the NLDAS domain

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -168,8 +168,11 @@ DATAMS = $(patsubst %, $(DSHARE_DIR)/%, $(SUMMA_DATAMS))
 # utility modules
 SUMMA_UTILMS= \
 		time_utils.f90 \
-		matrixOper.f90 \
-		mDecisions.f90
+		mDecisions.f90 \
+		snow_utils.f90 \
+		soil_utils.f90 \
+		updatState.f90 \
+		matrixOper.f90
 UTILMS = $(patsubst %, $(ENGINE_DIR)/%, $(SUMMA_UTILMS))
 
 # Model guts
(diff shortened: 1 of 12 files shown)


bartnijssen:
@martynpclark : Does this build OK on your end? For some reason the travis build failed on the `summa_install` part, but it is not clear to me why .... When you build locally, which compiler are you using?

martynpclark:
@bartnijssen -- I'm using `ifort` on `cheyenne`

martynpclark:
@bartnijssen -- do you know which compiler causes the failure? If so I can test locally.

bartnijssen:
I'm not even sure it's the compiler since there is no error message at all. It may just be that the compilation stalled. Let me kick it off one more time. On travis we are currently using gcc-6, which we need to update to 7 anyway, so may be that is the route to go before wasting your time.

bartnijssen:
@martynpclark : Compilation fails during compilation of `netcdf/read_icond.f90` with the message:

`Fatal Error: Can't open module file 'updatstate_module.mod'`

which is not built till later. Rerunning the compilation works (if you do not do a `make clean`), because the file does get created eventually.

The cleanest solution would be if we don’t need to call `updateSoil` from inside `read_icond` (in the same way that we do not call `updateSnow` from inside `read_icond`). In that case there is no dependency in the `netcdf` directory on things that are compiled later.

If that is not possible then we just need to reorg some of the directories a bit (or create the logic in the Makefile).

martynpclark:
@bartnijssen -- thanks. I forgot to update the Makefile. Now updated.

PR #301, opened 2018-01-10
Feature/output control

martynpclark:
Refactored the output control for summa (everything is backwards compatible). The reasons for this are two-fold:
1) The current file format is confusing;
2) The current approach uses timestep multiples, which makes it difficult to aggregate to time periods which are not a multiple of the time step (e.g., monthly or annual aggregation).

### DIFF
--- a/build/source/dshare/outpt_stat.f90
+++ b/build/source/dshare/outpt_stat.f90
@@ -156,6 +156,9 @@ contains
      tstat(iFreq) = -huge(tstat(iFreq))             !     - resets stat at beginning of period 
     case (iLookStat%mode)                           ! * mode over period (does not work)       
      tstat(iFreq) = realMissing
+    case default
+     message=trim(message)//'unable to identify type of statistic [reset]'
+     err=20; return
     ! -------------------------------------------------------------------------------------
    end select
   end if
@@ -184,6 +187,9 @@ contains
     if (tdata>tstat(iFreq)) tstat(iFreq) = tdata      !     - check value 
    case (iLookStat%mode)                              ! * mode over period (does not work)       
     tstat(iFreq) = realMissing
+   case default
+    message=trim(message)//'unable to identify type of statistic [calculating stats]'
+    err=20; return
    ! -------------------------------------------------------------------------------------
   end select
  end do ! looping through output frequencies
@@ -202,6 +208,7 @@ contains
     case (iLookStat%vari)                              ! * variance over period
      tstat(maxVarFreq+iFreq) = tstat(maxVarFreq+1)/statCounter(iFreq)            ! E[X] term
      tstat(iFreq) = tstat(iFreq)/statCounter(iFreq) - tstat(maxVarFreq+iFreq)**2 ! full variance
+    case default ! do nothing -- don't need finalization for most stats
     ! -------------------------------------------------------------------------------------
    end select
   end if
--- a/build/source/dshare/outpt_stat.f90
+++ b/build/source/dshare/outpt_stat.f90
@@ -20,29 +20,33 @@
 
 ! used to manage output statistics of the model and forcing variables
 module output_stats
-USE nrtype
+USE nrtype, realMissing=>nr_realMissing
+USE nrtype, integerMissing=>nr_integerMissing
 implicit none
 private
 public :: calcStats
-!public :: compileBasinStats
 contains
 
  ! ******************************************************************************************************
  ! public subroutine calcStats is called at every model timestep to update/store output statistics
  ! from model variables
  ! ******************************************************************************************************
- subroutine calcStats(stat,dat,meta,iStep,err,message)
+ subroutine calcStats(stat,dat,meta,resetStats,finalizeStats,statCounter,err,message)
  USE nrtype
  USE data_types,only:extended_info,dlength,ilength  ! metadata structure type
  USE var_lookup,only:iLookVarType                   ! named variables for variable types
  USE var_lookup,only:iLookStat                      ! named variables for output statistics types
  implicit none
 
- ! dummy variables
+ ! input variables
  type(dlength) ,intent(inout)   :: stat(:)          ! statistics
  class(*)      ,intent(in)      :: dat(:)           ! data
  type(extended_info),intent(in) :: meta(:)          ! metadata
- integer(i4b)  ,intent(in)      :: iStep            ! timestep index to compare with oFreq of each variable
+ logical(lgt)  ,intent(in)      :: resetStats(:)    ! vector of flags to reset statistics
+ logical(lgt)  ,intent(in)      :: finalizeStats(:) ! vector of flags to reset statistics
+ integer(i4b)  ,intent(in)      :: statCounter(:)   ! number of time steps in each output frequency
+
+ ! output variables
  integer(i4b)  ,intent(out)     :: err              ! error code
  character(*)  ,intent(out)     :: message          ! error message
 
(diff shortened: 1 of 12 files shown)


build/source/dshare/outpt_stat.f90
```diff
@@ -132,71 +137,75 @@ subroutine calc_stats(meta,stat,tdata,iStep,err,message)
...
+    case (iLookStat%mean)                           ! * mean over period                       
+     tstat(iFreq) = 0._dp                           !     - resets stat at beginning of period
+    case (iLookStat%vari)                           ! * variance over period                   
+     tstat(iFreq) = 0._dp                           !     - resets E[X^2] term in var calc    
+     tstat(maxVarFreq+iFreq) = 0._dp                !     - resets E[X]^2 term                 
+    case (iLookStat%mini)                           ! * minimum over period                    
+     tstat(iFreq) = huge(tstat(iFreq))              !     - resets stat at beginning of period 
+    case (iLookStat%maxi)                           ! * maximum over period                    
+     tstat(iFreq) = -huge(tstat(iFreq))             !     - resets stat at beginning of period 
+    case (iLookStat%mode)                           ! * mode over period (does not work)
```
bartnijssen: If `mode` does not work, do we need to include it? Or can we just strip it out of this PR?

build/source/dshare/outpt_stat.f90
```diff
@@ -132,71 +137,75 @@ subroutine calc_stats(meta,stat,tdata,iStep,err,message)
...
+   case (iLookStat%mean)                              ! * mean over period                       
+    tstat(iFreq) = tstat(iFreq) + tdata               !     -  increment data
+   case (iLookStat%vari)                              ! * variance over period                   
+    tstat(iFreq) = tstat(iFreq) + tdata**2                     ! - E[X^2] term in var calc    
+    tstat(maxVarFreq+iFreq) = tstat(maxVarFreq+iFreq) + tdata  ! - E[X]^2 term                 
+   case (iLookStat%mini)                              ! * minimum over period                    
+    if (tdata<tstat(iFreq)) tstat(iFreq) = tdata      !     - check value 
+   case (iLookStat%maxi)                              ! * maximum over period                    
+    if (tdata>tstat(iFreq)) tstat(iFreq) = tdata      !     - check value 
+   case (iLookStat%mode)                              ! * mode over period (does not work)
```
bartnijssen: Same as above

build/source/dshare/outpt_stat.f90
```diff
@@ -132,71 +137,75 @@ subroutine calc_stats(meta,stat,tdata,iStep,err,message)
...
+    case (iLookStat%vari)                           ! * variance over period                   
+     tstat(iFreq) = 0._dp                           !     - resets E[X^2] term in var calc    
+     tstat(maxVarFreq+iFreq) = 0._dp                !     - resets E[X]^2 term                 
+    case (iLookStat%mini)                           ! * minimum over period                    
+     tstat(iFreq) = huge(tstat(iFreq))              !     - resets stat at beginning of period 
+    case (iLookStat%maxi)                           ! * maximum over period                    
+     tstat(iFreq) = -huge(tstat(iFreq))             !     - resets stat at beginning of period 
+    case (iLookStat%mode)                           ! * mode over period (does not work)       
+     tstat(iFreq) = realMissing
+    ! -------------------------------------------------------------------------------------
```
bartnijssen: Can you add a default to trap any errors (it should never get there, but coding errors happen and this would be easier to catch them)?

build/source/dshare/outpt_stat.f90
```diff
@@ -132,71 +137,75 @@ subroutine calc_stats(meta,stat,tdata,iStep,err,message)
...
+   case (iLookStat%vari)                              ! * variance over period                   
+    tstat(iFreq) = tstat(iFreq) + tdata**2                     ! - E[X^2] term in var calc    
+    tstat(maxVarFreq+iFreq) = tstat(maxVarFreq+iFreq) + tdata  ! - E[X]^2 term                 
+   case (iLookStat%mini)                              ! * minimum over period                    
+    if (tdata<tstat(iFreq)) tstat(iFreq) = tdata      !     - check value 
+   case (iLookStat%maxi)                              ! * maximum over period                    
+    if (tdata>tstat(iFreq)) tstat(iFreq) = tdata      !     - check value 
+   case (iLookStat%mode)                              ! * mode over period (does not work)       
+    tstat(iFreq) = realMissing
+   ! -------------------------------------------------------------------------------------
```
bartnijssen: Can you add a default to trap any errors (it should never get there, but coding errors happen and this would be easier to catch them)?

build/source/dshare/outpt_stat.f90
```diff
@@ -132,71 +137,75 @@ subroutine calc_stats(meta,stat,tdata,iStep,err,message)
...
+   if(meta%statIndex(iFreq)==integerMissing) cycle     ! don't bother if output frequency is not desired for a given variable
+   if(meta%varType/=iLookVarType%outstat) cycle        ! only calculate stats for scalars
+   select case(meta%statIndex(iFreq))                  ! act depending on the statistic
+    ! -------------------------------------------------------------------------------------
+    case (iLookStat%mean)                              ! * mean over period
+     tstat(iFreq) = tstat(iFreq)/statCounter(iFreq)    !     - normalize sum into mean
+    case (iLookStat%vari)                              ! * variance over period
+     tstat(maxVarFreq+iFreq) = tstat(maxVarFreq+1)/statCounter(iFreq)            ! E[X] term
+     tstat(iFreq) = tstat(iFreq)/statCounter(iFreq) - tstat(maxVarFreq+iFreq)**2 ! full variance
+    ! -------------------------------------------------------------------------------------
```
bartnijssen: `default`

build/source/dshare/popMetadat.f90
```diff
@@ -751,188 +773,214 @@ subroutine read_output_file(err,message)
...
+    if(iFreq<0 .or. iFreq>maxvarFreq)then
+     message=trim(message)//'unable to identify desired output frequency for variable '//trim(varName)&
+                          //' [entered "'//trim(freqName)//'"]'
+     err=20; return
+    endif
+
+   ! temporally constant variables use timestep-level output (no aggregation)
+   case default
+    iFreq    = iLookFREQ%timestep
+    freqName = '[notNeeded; use timestep]'
```
bartnijssen: Odd name for a frequency. Why not assign `timestep`. May also be better to define this string in a module where all the other `freqName` options are defined rather than burying it deep in the code.

build/source/engine/read_force.f90
```diff
@@ -244,7 +245,8 @@ subroutine read_force(istep,iFile,iRead,ncid,time_data,forcStruct,err,message)
   end if
 
   ! convert julian day to time vector
-  call compcalday(dataJulDay,                     & ! input  = julian day
+  ! NOTE: use small offset to force ih=0 at the start of the day
```
bartnijssen: Does that work reliably?

build/source/netcdf/def_output.f90
```diff
@@ -326,60 +333,58 @@ subroutine def_variab(ncid,iFreq,spatialDesire,timeDesire,metaData,ivtype,err,me
...
+  catName = trim(metaData(iVar)%varunit)
+  if (iStat==iLookStat%totl) then
+
+   ! make sure that the units of this varaible allow for integration
+   if ((index(catName,'s-1')<=0).and.(index(catName,'s-2')<=0).and.(index(catName,'W m-2')<=0)) then
+    message=trim(message)//'trying to integrate a non-time variable: '//trim(metaData(iVar)%varName)//' - units: '//trim(catName)
+    err=20; return
+   endif
+
+   ! change to integrated units
```
bartnijssen: nice

build/source/dshare/outpt_stat.f90
martynpclark: Hmmmm.. this is something that @greyNearing added. Stripping it out now would require changing dozens of lines of code. Prefer to keep it for now.

build/source/dshare/outpt_stat.f90
martynpclark: Done

build/source/dshare/outpt_stat.f90
martynpclark: Done

build/source/dshare/outpt_stat.f90
martynpclark: Done

build/source/dshare/outpt_stat.f90
martynpclark: Done

build/source/dshare/popMetadat.f90
martynpclark: Fixed

build/source/engine/read_force.f90
martynpclark: Yes

build/source/netcdf/def_output.f90
martynpclark: @greyNearing did this -- just appears here because of a cut-and-paste job

bartnijssen: Still nice though to actually convert the units.

PR #305, opened 2018-01-12
Feature/isolate gru

martynpclark:
Refactor to isolate HRU computations. Tested before HRU refactor with identical results.

### DIFF
--- a/build/source/engine/run_oneGRU.f90
+++ b/build/source/engine/run_oneGRU.f90
@@ -24,45 +24,34 @@ module run_oneGRU_module
 USE nrtype
 
 ! access integers to define "yes" and "no"
-USE globalData,only:yes,no                 ! .true. and .false.
-
-! access the mapping betweeen GRUs and HRUs
-USE globalData,only:gru_struc              ! gru-hru mapping structures
+USE globalData,only:yes,no             ! .true. and .false.
 
 ! define data types
 USE data_types,only:&
+                    ! GRU-to-HRU mapping
+                    gru2hru_map,     & ! HRU info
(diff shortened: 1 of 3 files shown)


build/source/engine/run_oneGRU.f90
```diff
@@ -79,141 +68,142 @@ module run_oneGRU_module
  ! simulation for a single GRU
  subroutine run_oneGRU(&
                        ! model control
-                       iGRU,               & ! intent(in):    GRU index
```
bartnijssen: Do you want to pass the `iGRU` still into `run_oneGRU` simply for logging and error reporting purposes. In RASM we have had the problem in earlier versions that we would have an error in a low level routine and we wouldn't know where the error happened. If we get a complete stack trace when we report the error then that is not a problem. (note that it is a good sign that you do not need this info AT ALL to run the routine)

bartnijssen: Or is this part of `gruInfo`?

bartnijssen:
This series of PRs is really excellent for moving SUMMA forward. I know it makes no difference in the result, but it will enable much better scaling down the road!

build/source/engine/run_oneGRU.f90
martynpclark: We can get a traceback without including iGRU

bartnijssen: Perfect!

martynpclark: The routine will exit gracefully, and then IGRU will be available.

PR #307, opened 2018-01-14
further cleanup of module variables

martynpclark:
Further cleanup of module variables


martynpclark:
Tested against develop and results are identical.

PR #310, opened 2018-01-14
initial implementations of openMP

martynpclark:
This is the initial implementation of Noah-MP, thanks to extensive guidance from @gutmann. It is included here for initial code review. Do not merge.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -44,44 +44,70 @@
 # Define compiler flags. If you use a different compiler,
 # you will need to figure out what the equivalent flags are
 # and may need to update this section
+
+# ------------ define compiler flags ----------------------------------------
+
+# define open MP flags
+isOpenMP = yes
+FLAGS_OMP =
+LIBOPENMP =
+
+# Define compiler flags. If you use a different compiler,
+# you will need to figure out what the equivalent flags are
+# and may need to update this section
+
+# gfortran
 ifeq "$(FC)" "gfortran"
 
- # Define the flags for openMP
- FLAGS_OMP =
- #FLAGS_OMP = -fopenmp
+ifeq "$(isOpenMP)" "yes"
+ FLAGS_OMP = -fopenmp
+endif
 
- # Production run
- FLAGS_NOAH = -O3 -ffree-form -fdefault-real-8 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
- FLAGS_COMM = -O3 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
- FLAGS_SUMMA = -O3 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
+FLAGS_NOAH = -O3 -ffree-form -fdefault-real-8 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
+FLAGS_COMM = -O3 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
+FLAGS_SUMMA = -O3 -ffree-line-length-none -fmax-errors=0 $(FLAGS_OMP)
 
- # Debug run
- #FLAGS_NOAH = -p -g -ffree-form -fdefault-real-8 -ffree-line-length-none -fmax-errors=0 -fbacktrace -Wno-unused -Wno-unused-dummy-argument
- #FLAGS_COMM = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
- #FLAGS_SUMMA = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
+#FLAGS_NOAH = -p -g -ffree-form -fdefault-real-8 -ffree-line-length-none -fmax-errors=0 -fbacktrace -Wno-unused -Wno-unused-dummy-argument
+#FLAGS_COMM = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
+#FLAGS_SUMMA = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
 
 endif
+
+
+# ifort
 ifeq "$(FC)" "ifort"
 
- # Define the flags for openMP
- FLAGS_OMP =
- #FLAGS_OMP = -qopenmp
-
- # production run
- FLAGS_NOAH = -O3 -autodouble -noerror_limit -FR -auto -fltconsistency $(FLAGS_OMP)
- FLAGS_COMM = -O3 -FR -auto -fltconsistency -fpe0 $(FLAGS_OMP)
- FLAGS_SUMMA = -O3 -FR -auto -fltconsistency -fpe0 $(FLAGS_OMP)
-
- # debug run
- #FLAGS_NOAH = -O0 -p -g -warn nounused -autodouble -noerror_limit -FR -auto -WB -traceback -fltconsistency
- #FLAGS_COMM = -O0 -p -g -debug -warn all -check all -FR -auto -WB -traceback -fltconsistency -fpe0
- #FLAGS_SUMMA = -O0 -p -g -debug -warn all -check all -FR -auto -WB -traceback -fltconsistency -fpe0
-
+ifeq "$(isOpenMP)" "yes"
+ FLAGS_OMP = -qopenmp
 endif
+
+FLAGS_NOAH = -O3 -autodouble -noerror_limit -FR -auto -fltconsistency $(FLAGS_OMP)
+FLAGS_COMM = -O3 -FR -auto -fltconsistency -fpe0 $(FLAGS_OMP)
+FLAGS_SUMMA = -O3 -FR -auto -fltconsistency -fpe0 $(FLAGS_OMP)
+
+#FLAGS_NOAH = -O0 -p -g -warn nounused -autodouble -noerror_limit -FR -auto -WB -traceback -fltconsistency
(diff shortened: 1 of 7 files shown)


bartnijssen:
Need to make sure this runs and compiles with and without openmp

martynpclark:
@bartnijssen -- yes, this compiles and runs without openMP. The nice thing about the openMP directives is that they are interpreted as comments when the openMP flag is disabled.
```fortran
!$omp
```

build/source/driver/multi_driver.f90
```diff
@@ -827,32 +876,76 @@ program multi_driver
...
+ !call indexx(totalFluxCalls, ixExpense) ! ranking of each GRU w.r.t. computational expense
+ call indexx(timeGRU, ixExpense) ! ranking of each GRU w.r.t. computational expense
+ ixExpense=ixExpense(nGRU:1:-1)  ! reverse ranking: now largest to smallest
+
+ ! initialize the GRU count
+ ! NOTE: this needs to be outside the parallel section so it is not reinitialized by different threads
+ kGRU=0
+
+ ! initialize the time that the openMP section starts
+ openMPstart = omp_get_wtime()
```
jhamman: @martynpclark - this is a OpenMP function. Unless I'm missing something, this will require the OpenMP compiler flag, correct?

build/source/driver/multi_driver.f90
```diff
@@ -873,19 +966,39 @@ program multi_driver
...
+  !----- save timing information ------------------------------------------------
+
+  !$omp critical(saveTiming)
+
+  ! check errors
   call handle_err(err,message)
 
+  ! save timing information
+  timeGRU(iGRU)          = omp_get_wtime() - timeGRUstart(iGRU)
+  timeGRUcompleted(iGRU) = omp_get_wtime() - openMPstart
```
jhamman: nit. The use of spaces is inconsistent in this PR. I see lines with no space around `=`, aligned `=`, and 1 space on either side of ` = `. Does the SUMMA style guide specify how this should be?

build/source/driver/multi_driver.f90
martynpclark: Yes, I was wondering the same thing. The code compiles and runs fine without the `openMP` flag. Do you have suggestions on any `if..then..else` statements to wrap such things (e.g., if openMP then use the function, if not use system clock)?

build/source/noah-mp/module_sf_noahmplsm.F
```diff
@@ -63,6 +64,9 @@ module noahmp_globals
   REAL    :: RSMAX        !maximum stomatal resistance ( Assigned in REDPRM )
   REAL    :: TOPT         !optimum transpiration air temperature.
 
+  ! MPC change: make variables private for a given thread
+  !$omp threadprivate(NROOT, RGL, RSMIN, HS, RSMAX, TOPT)
+
```
jhamman: Suggest adding a stronger suggestion that at some point, these should be moved to a proper data structure.

build/source/driver/multi_driver.f90
jhamman: Not sure how this would work in Fortran per se but this is what I did in VIC: 

https://github.com/jhamman/VIC/blob/74dc13e77d613b811fa889e1b6339c8aa04a070e/vic/drivers/shared_image/include/vic_mpi.h#L32-L36

I use the system clocks.

bartnijssen:
@martynpclark : Didn't compile for me (or on travis). Complaining about missing functions associated with `openmp`

build/source/driver/multi_driver.f90
bartnijssen: No - https://github.com/NCAR/summa/blob/master/docs/development/SUMMA_coding_conventions.md

build/source/driver/multi_driver.f90
bartnijssen: It doesn't compile for me (`gfortran`)

build/Makefile
```diff
@@ -45,14 +45,38 @@
...
 ifeq "$(FC)" "gfortran"
- FLAGS_NOAH = -p -g -ffree-form -fdefault-real-8 -ffree-line-length-none -fmax-errors=0 -fbacktrace -Wno-unused -Wno-unused-dummy-argument
- FLAGS_COMM = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
- FLAGS_SUMMA = -p -g -Wall -ffree-line-length-none -fmax-errors=0 -fbacktrace -fcheck=bounds
+
+ # Define the flags for openMP
+ FLAGS_OMP =
+ #FLAGS_OMP = -fopenmp
+
+ # Production run
```
gutmann: Here (and for ifort and debug) can you set the FLAGS_NOAH to just be the flags needed for Noah, then have FLAGS_COMM be common(?) flags (O3,free, etc) and FLAGS_SUMMA be flags only used by summa (and not Noah?)

build/source/driver/multi_driver.f90
```diff
@@ -276,7 +283,21 @@ program multi_driver
 integer(i4b)                     :: fileHRU                    ! number of HRUs in the input file
 integer(i4b)                     :: iRunMode                   ! define the current running mode
 character(len=128)               :: fmtGruOutput               ! a format string used to write start and end GRU in output file names
- ! version information generated during compiling
+! timing information
```
gutmann: along the lines of Joe's comment, this could all be abstracted into a timing module.  I have a timer object implemented here if you are interested: 
https://github.com/gutmann/coarray_icar/blob/develop/src/objects/timer_interface.f90
https://github.com/gutmann/coarray_icar/blob/develop/src/objects/timer_implementation.f90
then you just have a timeGRU timer object, and call timeGRU%start() and timeGRU%stop(), timeGRU%get_time() (or as_string())

martynpclark:
@bartnijssen -- updated driver to remove any openMP dependencies.

bartnijssen:
@martynpclark : I also need to include the `-lgomp` directive during linking to get the model to compile in case I specify `FLAGS_OMP = -fopenmp`. Let me know if you want me to update the Makefile and issue a PR against yours.

bartnijssen:
I also should probably update the travis file to compile with and without openmp. (Currently we are only using travis to make sure that the model compiles - we are not doing any actual test simulations).

jhamman:
@martynpclark - One last comment here. Do you want to update the SUMMA installation documentation as part of this PR? I think it could probably use a section on OpenMP parallelization, setting the `OMP_NUM_THREADS` variable, etc.

build/source/driver/multi_driver.f90
```diff
@@ -827,32 +874,75 @@ program multi_driver
...
+
+  ! assign expensive GRUs to threads that enter first
+  kGRU=kGRU+1
+  iGRU=ixExpense(kGRU)
+
+  ! get the time that the GRU started
+  call system_clock( timeGRUstart(iGRU) )
+
+  ! print progress
+  !write(*,'(a,1x,5(i4,1x),f20.10,1x)') 'iGRU, jGRU, kGRU, nThreads = ', iGRU, jGRU, kGRU, nThreads, timeGRU(iGRU)
```
gutmann: Do you want to remove debugging related comments?

build/source/driver/multi_driver.f90
```diff
@@ -189,7 +190,7 @@ program multi_driver
 type(gru_hru_double)             :: dparStruct                 ! x%gru(:)%hru(:)%var(:)     -- default model parameters
 ! define indices
 integer(i4b)                     :: iStruct                    ! loop through data structures
-integer(i4b)                     :: iGRU,jGRU                  ! index of grouped response unit
+integer(i4b)                     :: iGRU,jGRU,kGRU             ! index of grouped response unit
```
gutmann: Can i,j,k be given more meaningful names?

gutmann: or at least some comment here on what they are used for?  I tend to think of i,j,k as being nested loops which isn't really what you are doing (from what I can see? )

build/source/driver/multi_driver.f90
```diff
@@ -264,8 +265,12 @@ program multi_driver
 character(len=256)               :: timeString                 ! protion of restart file name that contains the write-out time
 character(len=256)               :: restartFile                ! restart file name
 character(len=256)               :: attrFile                   ! attributes file name
+! open MP functions
+integer(i4b)                     :: omp_get_thread_num,omp_get_num_threads
```
gutmann: Are these still used or can they be removed?

build/source/noah-mp/module_sf_noahmplsm.F
martynpclark: Good point. This suggestion will be made as part of CTSM development.
https://github.com/ESCOMP/ctsm

build/source/driver/multi_driver.f90
martynpclark: Yes, this would be nice in principle. I will not address this as part of the current PR.

martynpclark:
@bartnijssen -- Makefile updated to include openMP options.

build/source/driver/multi_driver.f90
martynpclark: They are used to index different elements of the GRU structure (as stated). This is none in different ways in different parts of the code, so adding more meaningful names is not possible without adding more variables. Will retain as is.

build/source/driver/multi_driver.f90
martynpclark: The function `omp_get_num_threads` is still used. This has been retained and the other function names have been removed.

Note that we use the code
```fortran
nThreads = 1
 !$ nThreads = omp_get_num_threads()
```

and the line `!$ nThreads = omp_get_num_threads()` is interpreted as a comment if the `openMP` libraries are not linked.

martynpclark:
@gutmann -- new commit in response to code review.

PR #311, opened 2018-01-15
Feature/refactor forcing read

martynpclark:
Refactor read_force to ease optimization

### DIFF
--- a/build/source/dshare/globalData.f90
+++ b/build/source/dshare/globalData.f90
@@ -174,6 +174,7 @@ MODULE globalData
  real(dp),save,public                        :: fracJulday              ! fractional julian days since the start of year
  real(dp),save,public                        :: dJulianStart            ! julian day of start time of simulation
  real(dp),save,public                        :: dJulianFinsh            ! julian day of end time of simulation
+ integer(i4b),save,public                    :: nHRUfile                ! number of HRUs in the file
  integer(i4b),save,public                    :: yearLength              ! number of days in the current year
  integer(i4b),save,public                    :: urbanVegCategory        ! vegetation category for urban areas
  logical(lgt),save,public                    :: doJacobian=.false.      ! flag to compute the Jacobian
--- a/build/source/dshare/globalData.f90
+++ b/build/source/dshare/globalData.f90
@@ -170,6 +170,7 @@ MODULE globalData
  integer(i4b),save,public                    :: numtim                  ! number of time steps
  real(dp),save,public                        :: data_step               ! time step of the data
  real(dp),save,public                        :: refJulday               ! reference time in fractional julian days
+ real(dp),save,public                        :: refJulday_data          ! reference time in fractional julian days (data files)
  real(dp),save,public                        :: fracJulday              ! fractional julian days since the start of year
  real(dp),save,public                        :: dJulianStart            ! julian day of start time of simulation
  real(dp),save,public                        :: dJulianFinsh            ! julian day of end time of simulation
--- a/build/source/engine/ffile_info.f90
+++ b/build/source/engine/ffile_info.f90
@@ -153,6 +153,16 @@ contains
   ! set nVar attribute
   forcFileInfo(iFile)%nVars = nVar
 
+  ! allocate space
+  allocate(forcFileInfo(iFile)%var_ix(nVar), stat=err)
+  if(err/=0)then
+   message=trim(message)//'problem allocating space for structure element'
+   err=20; return
+  endif
+
+  ! initialize data structure
+  forcFileInfo(iFile)%var_ix(:) = integerMissing
+
   ! inquire nhru dimension size
   err = nf90_inq_dimid(ncid,'hru',dimId);                 if(err/=0)then; message=trim(message)//'cannot find dimension hru'; return; endif
   err = nf90_inquire_dimension(ncid,dimId,len=file_nHRU); if(err/=0)then; message=trim(message)//'cannot read dimension hru'; return; endif
@@ -164,7 +174,7 @@ contains
   ! loop through all variables in netcdf file, check to see if everything needed to run the model exists and data_step is correct
   do iNC=1,nVar
 
-   ! inqure about current variable name, type, number of dimensions
+   ! inquire about current variable name, type, number of dimensions
    err = nf90_inquire_variable(ncid,iNC,name=varName)
    if(err/=0)then; message=trim(message)//'problem inquiring variable: '//trim(varName); return; end if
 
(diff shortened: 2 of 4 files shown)


build/source/dshare/globalData.f90
```diff
@@ -170,9 +170,11 @@ MODULE globalData
  integer(i4b),save,public                    :: numtim                  ! number of time steps
  real(dp),save,public                        :: data_step               ! time step of the data
  real(dp),save,public                        :: refJulday               ! reference time in fractional julian days
+ real(dp),save,public                        :: refJulday_data          ! reference time in fractional julian days (data files)
```
bartnijssen: What's the difference between with `refJulday` and `refJulday_data`?

build/source/engine/ffile_info.f90
```diff
@@ -164,7 +174,7 @@ subroutine ffile_info(nGRU,err,message)
   ! loop through all variables in netcdf file, check to see if everything needed to run the model exists and data_step is correct
   do iNC=1,nVar
```
bartnijssen: Should the loop be over the number of variables in the file or over the number of variables that you already know that you need? Seems to me I could really slow SUMMA down if I give it a file with a huge number of variables that it would not need. Flipping the loop (i.e. specifically looking for the variables that you actually know you need) may be more of a fixed time.

build/source/engine/read_force.f90
```diff
@@ -20,23 +20,51 @@
...
+USE summaFileManager,only:INPUT_PATH          ! path of the forcing data file
+
+! privacy
 implicit none
 private
 public::read_force
 
+! global parameters
+real(dp),parameter  :: verySmall=1e-3_dp      ! tiny number
+real(dp),parameter  :: smallOffset=1.e-8_dp   ! small offset (units=days) to force ih=0 at the start of the day
```
bartnijssen: Should these be defined globally?

build/source/engine/read_force.f90
```diff
@@ -417,49 +214,313 @@ subroutine read_force(istep,iFile,iRead,ncid,time_data,forcStruct,err,message)
...
+   if(time_data(iline)==integerMissing)then; err=40; message=trim(message)//"variableMissing[var='"//trim(time_meta(iline)%varname)//"']"; return; end if
+  end do
+ end if
+
+ ! initialize flags for forcing data
+ checkForce(:) = .false.
+ checkForce(iLookFORCE%time) = .true.  ! time is handled separately
+
+ ! loop through forcing data variables
+ do iNC=1,forcFileInfo(iFile)%nVars
```
bartnijssen: Same as above - would it be faster if you only look for the data that you know you need? Rather than checking all the variables in the file?

build/source/engine/read_force.f90
```diff
@@ -417,49 +214,313 @@ subroutine read_force(istep,iFile,iRead,ncid,time_data,forcStruct,err,message)
...
+    endif
+
+    ! check the number of HRUs
+    if(iHRU_global > nHRUfile)then
+     message=trim(message)//'HRU index exceeds the number of HRUs in the forcing data file'
+     err=20; return
+    endif
+
+    ! get individual data value
+    if(simultaneousRead) dataVal(1) = dataVec(iHRU_local)
```
bartnijssen: Isn't this to get the forcing data for all HRUs? In that case: fix comment.

bartnijssen: If this is to get the data for all HRUs, shouldn;t this be outside the HRU loop, i.e.:

```Fortran
do iGRU=1,size(gru_struc)
  if(simultaneousRead) ..... 
  else !loop over the HRUs and do individual reads
    do iHRU=1,gru_struc(iGRU)%hruCount .....
    end do
  end if
end do
```

bartnijssen: This is still only getting one time step per read - is that correct?

build/source/dshare/globalData.f90
martynpclark: `refJulday` is the reference time for the model simulation and `refJulday_data` is the reference time in the data files (as noted in the comments). These can be different.

build/source/engine/ffile_info.f90
martynpclark: We could do it either way. We loop over the number of variables in the data file in case there are efficiency gains by reading data close together on disk. Note that `ffile_info.f90` is only called once at the start of the simulation.

build/source/engine/read_force.f90
martynpclark: These parameters are specific to `read_force.f90`. We have similar parameters in other routines (set to different values) which are also local in scope.

build/source/engine/read_force.f90
martynpclark: We only read the data that we need.

martynpclark: The very next line of code skips variables that are not needed:
```fortran
! loop through forcing data variables
 do iNC=1,forcFileInfo(iFile)%nVars

  ! check variable is desired
  if(forcFileInfo(iFile)%var_ix(iNC)==integerMissing) cycle
```

martynpclark: It could be even more clear if we combined these statements into one line in a `do concurrent` but I need to check that there are no dependencies among different loop iterations. Leaving as is for now.

```fortran
do concurrent(iNC=1:forcFileInfo(iFile)%nVars, forcFileInfo(iFile)%var_ix(iNC)/=integerMissing)
```

build/source/engine/read_force.f90
martynpclark: This is getting one time step per read. We get all HRUs at once if `simultaneousRead==.true.`

martynpclark: The code on line 496: This extracts an individual data value from the data vector, so the comment is correct.

The logic is as follows:

```fortran
  ! read forcing data for all HRUs
  if(simultaneousRead)then
   err=nf90_get_var(ncid,forcFileInfo(iFile)%data_id(ivar),dataVec,start=(/ixHRUfile_min,iRead/),count=(/nHRUlocal,1/))
   if(err/=nf90_noerr)then; message=trim(message)//'problem reading forcing data: '//trim(varName)//'/'//trim(nf90_strerror(err)); return; endif
  endif
```

```fortran
  ! loop through GRUs and HRUs
  do iGRU=1,size(gru_struc)
   do iHRU=1,gru_struc(iGRU)%hruCount
```

```fortran
! get individual data value
if(simultaneousRead) dataVal(1) = dataVec(iHRU_local)
```

```fortran
! put the data into structures
forcStruct%gru(iGRU)%hru(iHRU)%var(ivar) = dataVal(1)
```

martynpclark: So: The first code block reads all HRUs (populate `dataVec`) the third code block copies the desired part of `dataVec` into `dataVal` and the fourth code block puts `dataVal` into the data structure.

We could have also written
```fortran
forcStruct%gru(iGRU)%hru(iHRU)%var(ivar) = dataVec(iHRU_local)
```

but I copied `dataVec` into `dataVal` so that I could use the same code to operate on `dataVal` when `simultaneousRead==.false.`

PR #314, opened 2018-01-25
adding tracking to git repository

julievano:
this 2 lines of tracking code - i included an icon, but could change this if so desired

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,3 +36,6 @@ In addition, an NCAR technical note describes the SUMMA implementation in detail
 
 ## License
 SUMMA is distributed under the GNU Public License Version 3. For details see the file `COPYING` in the SUMMA root directory or visit the [online version](http://www.gnu.org/licenses/gpl-3.0.html).
+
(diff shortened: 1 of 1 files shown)


julievano:
@bartnijssen - yes, they will be counted (whenever the image appears, it sends a note to google analytics).  I've thought about this a little - but would welcome others suggestions.  in counting these visits, we are counting every time someone is at there github repo (either forked or the main one).  which, if our goal is to understand how much people are viewing the code, works.  As we move forward though, I was going to compare this with other tracking that is more direct... we can always change the tracking number and find another way if this becomes a problem.

bartnijssen:
I think this requires a bit more discussion. I am concerned with tracking access to people's private forks of the repo, since this is not behavior that they may expect. Given privacy concerns in general, I don't want people to refrain from forking SUMMA because their access to their own repo will be tracked by us. In consultation with @julievano I am closing this for now pending some more discussion about this with @martynpclark and others. We can easily reopen this PR after that discussion as need be.

PR #299, opened 2017-12-19
DO NOT MERGE: big bucket aquifer parameterization

martynpclark:
DO NOT MERGE: still testing

Implementation of the big bucket aquifer parameterization

PR #317, opened 2018-01-30
Feature/hru aquifer

martynpclark:
Single aquifer per hru, updated for consistency with latest develop branch


martynpclark:
Requires testing from @msaharia

msaharia:
There is a discrepancy in values from the old implementation and new. More info is in the notebooks:

### Case 1 (old SUMMA implementation):  `2650057` + #299

 SUM
pptrate (mm/h) : 4094.2294403329893
averageInstantRunoff (mm/h) : 6071.426016334867
scalarRainPlusMelt (mm/h) : 2567.3218490895383
scalarAquiferRecharge (mm/h) : 3035.713176155699
scalarAquiferBaseflow (mm/h) : 3035.712840179139
scalarAquiferStorage (mm) : 2575.28424754777
scalarAquiferTranspire (mm/h) : 0.0

https://github.com/msaharia/luit/blob/master/summatesting_dec7.ipynb

### Case 2: (new SUMMA implementation):  `7604c31` + #317 

SUM
pptrate (mm/h) : 4094.2294403329893
averageInstantRunoff (mm/h) : 7815.9517905182
scalarRainPlusMelt (mm/h) : 3589.4263034175215
scalarAquiferRecharge (mm/h) : 3907.9612114387346
scalarAquiferBaseflow (mm/h) : 3907.990579079458
scalarAquiferStorage (mm) : 3366.955246483139
scalarAquiferTranspire (mm/h) : 0.0

https://github.com/msaharia/luit/blob/master/summatesting_jan28.ipynb

TBD

bartnijssen:
@msaharia : Nice use of the jupyter notebooks!

That snow record looks weird in the new simulations. Is there a way to check that the temperatures that are "seen" by the HRU are actually the same (I understand they should be - the question is whether they are). In the first few years in the new simulation you never actually melt the entire snow pack for this HRU.

build/source/engine/getVectorz.f90
```diff
@@ -528,6 +559,9 @@ subroutine varExtract(&
   end do  ! looping through non-missing energy state variables in the snow+soil domain
  endif
 
+ ! extract temperature of the canopy air space
```
bartnijssen: Update comment - I don't think you're extracting the `temperature of the canopy air space` here

build/source/dshare/globalData.f90
```diff
@@ -124,6 +124,7 @@ MODULE globalData
  integer(i4b),parameter,public               :: iname_veg =1001         ! named variable to denote a vegetation state variable
  integer(i4b),parameter,public               :: iname_soil=1002         ! named variable to denote a soil layer
  integer(i4b),parameter,public               :: iname_snow=1003         ! named variable to denote a snow layer
+ integer(i4b),parameter,public               :: iname_aquifer=1004      ! named variable to denote a snow layer
```
bartnijssen: Update comment: `aquifer layer`, not `snow layer`

msaharia:
@bartnijssen @martynpclark Right. I outputted the `mLayerTemp` and plotted the top layer.

Reference: `mLayerTemp(1)        = min(maxFrozenSnowTemp,surfaceLayerSoilTemp)    ! snow temperature  (K)` in `source/engine/layerDivide.f90`

The notebooks have been updated for context. 

old

new

bartnijssen:
Is there a way to save out the actual air temperature that that HRU experiences? The layer temperature you are showing is consistent with the swe plot (temperature of the top layer does not exceed 0 when there is snow), but I don't think it is the cause of it (just another expression of the same thing). Has there been a change in the inputs between these two code versions - so that something is misinterpreted?)

bartnijssen:
@msaharia : Manab - are you testing the right version: 7604c31 is **_before_** the edits made as part of this PR. Can you checkout this PR locally and re-run? You should be testing everything up to and including f575849

To checkout locally:

```
git checkout -b martynpclark-feature/hruAquifer develop
git pull https://github.com/martynpclark/summa.git feature/hruAquifer
```

msaharia:
Air temperatures look the same. Notebooks above have been updated.

#Old 
old

#New
new

Note: The exact same input files from forcing to parameters are used in both.

msaharia:
@bartnijssen Yes, I am on `f575849`. In Case 2, when I mentioned `7604c31 + #317`, I meant I am on the latest commit on develop + PR. Sorry for the confusion.

msaharia:
Further Diagnosis. As per offline discussions with @martynpclark, I have checked out PRs systematically between those times  (Without BigBuckt, ofcourse) and, so far, I have narrowed it down to some change between b3bdcbb1b1b65c8bd7e3bb093d2d37e87b340979 and 26500576f77034adfbce43b153fd1c97e5af6962

Notebook for b3bdcbb1b1b65c8bd7e3bb093d2d37e87b340979: https://github.com/msaharia/luit/blob/master/summatesting_b3bdcbb.ipynb
Notebook for 26500576f77034adfbce43b153fd1c97e5af6962: https://github.com/msaharia/luit/blob/master/summatesting_2650057.ipynb

I shall have more once I figure out which commit the change occurred in exactly.

bartnijssen:
Here is the diff between these two commits: https://github.com/NCAR/summa/compare/26500576f77034adfbce43b153fd1c97e5af6962...b3bdcbb1b1b65c8bd7e3bb093d2d37e87b340979

bartnijssen:
The biggest change related to snow is that snowfall is added to the snowpack _**AFTER**_  we run the main solver rather than before (@martynpclark : can you explain why that was done). My guess is that with the move, the `newsnwfall()` routine either did not end up in the right place, or that need to update `nSnow` before we enter the `newsnwfall()` routine (after the loop through the substeps). So adding:

```
nSnow = count(indx_data%var(iLookINDEX%layerType)%dat==iname_snow)
 nSoil = count(indx_data%var(iLookINDEX%layerType)%dat==iname_soil)
```
before entering `newsnwfall()` may do the trick.

@martynpclark : This may be immediately obvious to you when you take a look.

martynpclark:
@bartnijssen @msaharia -- thanks for pinning this down. I agree that the major change here is moving the `newSnwfall` routine to after the solver.

This was done because the state vector is defined by the call to `vegPhenlgy` at the start of the time step. Specifically, the call to `vegPhenlgy` computes the exposure of vegetation (if vegetation is buried by snow), and if the vegetation state variables need to be computed. 

I encountered a rare situation where vegetation was buried by new snow that fell within the time step, so I moved the `newSnwfall` routine to after the solver.

Note that the `newSnwfall` routine cannot be before the call to `vegPhenlgy`  because unloading of snow from the canopy is included as part of the snowfall, and `newSnwfall` needs to be called after `canopySnow`.

bartnijssen:
@martynpclark : Just curious -- is there a reason for the two different ways of specifying `nsnow`, `nsoil` and `nlayers` before and after `coupled_em.f90:volicepack()` https://github.com/martynpclark/summa/blob/f5758495d69aa4dd3e5520cc99f82067e5c67795/build/source/engine/coupled_em.f90#L936-L991?

bartnijssen:
Just confirming that I can reproduce the behavior observed by Manab with `wrrPaperTestCases/figure06/`

image
(numbers in legend refer to commits)

bartnijssen:
I think I isolated the problem to a single commit and not the one associated with moving the snowfall:

The commit is this one: https://github.com/NCAR/summa/compare/4cb92cd...fbebb

The jump in SWE occurs after this commit is made:

image

I'll look a bit further - but my guess is that the shortwave calculations end up being done with the wrong time or something and thus the snow is not melting

martynpclark:
@bartnijssen -- you found it!

martynpclark:
I meant to go back and check this. All forcing files should be in UTC now.

martynpclark:
I expect that if you change the forcing data to UTC then results will be identical.

bartnijssen:
Yep - setting  `timeOffset = 0._dp` in `derivforce.f90` solves the problem. We probably just need a switch for whether the time is local or UTC.

bartnijssen:
@msaharia : You can also just change all your forcings to UTC (simple shift of the time axis). That will work with the version of the code that you have. We should debate whether we will allow both and have a switch to calculate `timeOffset`. The latter may be a bit more user friendly for users that do point simulations.

bartnijssen:
Note that the time issue is separate from this PR. See issue #318.

@msaharia : Let me know if this can be merged. We'll address the `timeOffset` issue separately.

PR #324, opened 2018-02-27
Feature/fast reading

martynpclark:
Speed up reading of input data.

### DIFF
--- a/build/source/driver/multi_driver.f90
+++ b/build/source/driver/multi_driver.f90
@@ -116,6 +116,7 @@ USE globalData,only:localParFallback                        ! local column defau
 USE globalData,only:basinParFallback                        ! basin-average default parameters
 USE globalData,only:structInfo                              ! information on the data structures
 USE globalData,only:numtim                                  ! number of time steps
+USE globalData,only:chunksize                               ! chunk size for netcdf read/write
 USE globalData,only:urbanVegCategory                        ! vegetation category for urban areas
 USE globalData,only:greenVegFrac_monthly                    ! fraction of green vegetation in each month (0-1)
 USE globalData,only:globalPrintFlag                         ! global print flag
@@ -422,6 +423,9 @@ call handle_err(err,message)
 
 ! get the number of HRUs in the run domain
 nHRUrun = sum(gru_struc%hruCount)
+
+! define the chunksize (for netcdf read/write)
+chunksize=8*nHRUrun
 
 ! *****************************************************************************
 ! *** read description of model forcing datafile used in each HRU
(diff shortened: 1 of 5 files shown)


build/source/driver/multi_driver.f90
```diff
@@ -423,6 +424,9 @@ program multi_driver
 ! get the number of HRUs in the run domain
 nHRUrun = sum(gru_struc%hruCount)
 
+! define the chunksize (for netcdf read/write)
+chunksize=8*nHRUrun
```
gutmann: I suspect that ~2x nHRUrun would be better, but you'd have to run some tests on this.  The files themselves should probably have larger chunk sizes to accommodate runs with larger numbers of HRUs, but once you know exactly how many HRUs you have you might even do better setting `chunksize = nHRUrun`.  The writes you might want to keep larger so that other applications working with more data at once will work well, though if you are writing one file / process, then `chunksize = nHRUrun` might be better there too.

build/source/engine/ffile_info.f90
```diff
@@ -153,15 +147,19 @@ subroutine ffile_info(nGRU,err,message)
   ! set nVar attribute
   forcFileInfo(iFile)%nVars = nVar
 
-  ! allocate space
-  allocate(forcFileInfo(iFile)%var_ix(nVar), stat=err)
-  if(err/=0)then
-   message=trim(message)//'problem allocating space for structure element'
-   err=20; return
-  endif
+  ! ensure allocatable structure components are deallocated
```
gutmann: Is there a reason to include all of this in the "read faster" pull request?

build/source/driver/multi_driver.f90
martynpclark: This is 2x nHRUrun -- since most of the forcing data is stored as 4-byte reals.

martynpclark: I should really have `8` be a parameter

build/source/engine/ffile_info.f90
bartnijssen: Agreed - can we split this so the chunksize PR is separate and its own thing? Much easier to track if we need to modify it later (or roll it back).

bartnijssen: It should be easy to split the two commits into separate PRs.

bartnijssen:
@martynpclark : Can you split the `chunksize` into a separate PR? Then I can go ahead and merge and include in SUMMA 2.1

arbennett:
This was split into #340 and #341. Closing.

PR #329, opened 2018-04-12
Set timeOffset to 0.

arbennett:


bartnijssen:
@arbennett : Can you check to make sure this gives the right answer with the test files that Choi has

PR #335, opened 2018-10-11
Feature/lis integration

martynpclark:
Refactor of the lis driver


arbennett:
It appears that there is some sort of bug when using LAI method of `specified` that causes large changes in snowpack and turbulent heat fluxes. @martynpclark do you have any guesses on why this would happen?

arbennett:
This is superseded by #343. Closing.

Issue #338, opened 2019-05-24
Crash when trying to divide snow layers above maxSnowLayers if using jrdn1991 layering


arbennett:
SUMMA will segfault due to trying to access out of bounds memory if the number of snow layers tries to go above `maxSnowLayers`. This only occurs while using the `jrdn1991` layering scheme. The main issue is that there aren't adequate checks if the maximum number of snow layers has been reached.

arbennett:
Solved via #345

PR #345, opened 2019-08-08
Bugfix/max snow layers

arbennett:
This pull request provides a fix for #338. It is bit for bit with the develop branch.

### DIFF
--- a/build/source/engine/layerDivide.f90
+++ b/build/source/engine/layerDivide.f90
@@ -31,7 +31,6 @@ USE multiconst,only:&
 ! access named variables for snow and soil
 USE globalData,only:iname_snow        ! named variables for snow
 USE globalData,only:iname_soil        ! named variables for soil
-USE globalData,only:maxSnowLayers     ! maximum number of snow layers
 
 ! access missing values
 USE globalData,only:integerMissing  ! missing integer
@@ -95,6 +94,7 @@ contains
  ! --------------------------------------------------------------------------------------------------------
  ! computational modules
  USE snow_utils_module,only:fracliquid,templiquid              ! functions to compute temperature/liquid water
+ USE globalData,only:maxSnowLayers     ! maximum number of snow layers
  implicit none
  ! --------------------------------------------------------------------------------------------------------
  ! input/output: model data structures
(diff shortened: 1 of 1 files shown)


build/source/engine/layerDivide.f90
```diff
@@ -259,17 +260,24 @@ subroutine layerDivide(&
 
   ! identify the number of layers to check for need for sub-division
   select case(ix_snowLayers)
-   case(sameRulesAllLayers);    nCheck = nSnow
-   case(rulesDependLayerIndex); nCheck = min(nSnow,4)  ! the depth of the 5th layer, if it exists, does not have a maximum value
+   case(sameRulesAllLayers);    nCheck = min(nSnow, maxSnowLayers-1)
+   case(rulesDependLayerIndex); nCheck = min(nSnow, maxSnowLayers-1)  ! the depth of the 5th layer, if it exists, does not have a maximum value
```
bartnijssen: Fix comment to: `the depth of the last layer (maxSnowLayers), if it exists` ...

build/source/engine/layerDivide.f90
```diff
@@ -259,17 +260,24 @@ subroutine layerDivide(&
...
 
   ! loop through all layers, and sub-divide a given layer, if necessary
   do iLayer=1,nCheck
+   divideLayer=.false.
 
    ! identify the maximum depth of the layer
    select case(ix_snowLayers)
-    case(sameRulesAllLayers);    zmaxCheck = zmax
+    case(sameRulesAllLayers)
+     if (nCheck >= 99) then
```
bartnijssen: What are the magic numbers `99` and two lines laters `99999.`?

bartnijssen:
@arbennett : Can you just give quick answers to the review questions? I don't quite know what the 99 was for (i understand the 99999. --> veryBig)

build/source/engine/layerDivide.f90
arbennett: 99 was a standin for `maxSnowLayers-1`. The `jrdn1991` snow layering decision sets `maxSnowLayers` to 100. So, the aim is to never split the snowpack into more than `maxSnowLayers`.

PR #375, opened 2020-03-19
updates scalarTotalRunoff calculation in case of aquifer bigbucket

andywood:
I also included a change of default netcdf formatting that I thought we landed on after all the AIST discussion.  Please confirm that this is something we should leave in.  Rather than NF90_HDF5 the default would be NF90_NETCDF4.


build/source/engine/computFlux.f90
```diff
@@ -793,6 +793,9 @@ subroutine computFlux(&
                    err,cmessage)                   ! intent(out): error control
    if(err/=0)then; message=trim(message)//trim(cmessage); return; endif
 
+   ! compute total runoff (overwrite previously calculated value before considering aquifer)
+   scalarTotalRunoff  = scalarSurfaceRunoff + scalarAquiferBaseflow
+
```
martynpclark: Is this double counting soil drainage? Note that `scalarSoilDrainage` is an input to the `bigAquifer` and is included in the total runoff calculation earlier

build/source/engine/updateVars.f90
```diff
@@ -628,9 +628,14 @@ subroutine updateVars(&
...
+   ! check liquid water (include tolerance)
+   if(mLayerVolFracLiqTrial(iLayer) > theta_sat(ixControlIndex)+epsT )then
     message=trim(message)//'liquid water greater than porosity'
+    print*,'---------------'
+    print*,'porosity(theta_sat)=', theta_sat(ixControlIndex)
+    print*,'liq water =',mLayerVolFracLiqTrial(iLayer)
+    print*,'layer =',iLayer
+    print*,'---------------'
     err=20; return
    endif
```
martynpclark: Does this have anything to do with the runoff calculation?

build/source/netcdf/def_output.f90
```diff
@@ -125,7 +125,7 @@ subroutine def_output(summaVersion,buildTime,gitBranch,gitHash,nGRU,nHRU,nSoil,i
   fname   = trim(infile)//'_'//trim(fstring)//'.nc'
   call ini_create(nGRU,nHRU,nSoil,trim(fname),ncid(iFreq),err,cmessage)
   if(err/=0)then; message=trim(message)//trim(cmessage); return; end if
-  print*,'Created output file: '//trim(fname)//'; ncid = ', ncid(iFreq)
+  print*,'Created output file: '//trim(fname)//'   !; ncid = ', ncid(iFreq)
```
martynpclark: I can see that this is a bug (semicolon should be a comma? and no double dash?). But a separate PR?

build/source/netcdf/def_output.f90
```diff
@@ -212,7 +212,8 @@ subroutine ini_create(nGRU,nHRU,nSoil,infile,ncid,err,message)
 
  ! create output file
  !err = nf90_create(trim(infile),NF90_64BIT_OFFSET,ncid)
- err = nf90_create(trim(infile),NF90_HDF5,ncid)
+ !err = nf90_create(trim(infile),NF90_HDF5,ncid)
+ err = nf90_create(trim(infile),NF90_NETCDF4,ncid)
```
martynpclark: why NETCDF4? Another PR?

build/source/netcdf/def_output.f90
```diff
@@ -348,7 +349,7 @@ subroutine def_variab(ncid,iFreq,spatialDesire,timeDesire,metaData,ivtype,err,me
   ! define statistics index
   iStat = metaData(iVar)%statIndex(iFreq)
 
-  ! create full variable name (append statistics info(
+  ! create full variable name (append statistics info)
```
martynpclark: OK

build/source/engine/computFlux.f90
andywood: I don't think so. This should just overwrite the earlier assignment.  In one case I tested, scalarRO decreased because the aquifer was delaying runoff, versus a case in which scalar RO included soil drainage but was being immediately expelled from the aquifer (aquiferBFrate = 0.1).

build/source/engine/updateVars.f90
andywood: no, this was an earlier change I never added, meant to solve a case where the tolerances were stopping the run, and to provide some info.  If there's too much info (cluttering the code), the added print statements can be removed.  The epsT did help, though.

build/source/netcdf/def_output.f90
andywood: yep, I wish I could say I put that there just to check that people were paying attention but the ' after // is a mistake.  The idea was to remove the ncid info which is informative if you happen to be in a debug mode on ncids, but not something a typical user would need to see all the time.

build/source/netcdf/def_output.f90
andywood: This is worth discussing again, perhaps.  There was a long interaction on it in AIST because NASA was having trouble with the NF90_HDF5, which is apparently also now a deprecated mode.  The NF90_64BIT_OFFSET was motivated to handle large files, but I think does not handle int64s.  Naoki also discussed with Eric K and his advice was to stick with NETCDF4 as being more solid.  If we're offering a default, this seemed to be the one.  I've been using this with all the runs since the NASA discussions.  Others probably know more than I but I didn't want this to get lost.  Perhaps raising the other two modes as options in a brief code comment is a way to give a safe option and let users 'upgrade' if their needs require it.

build/source/netcdf/def_output.f90
andywood: I can correct this with another PR - I notice that when you're not running in hourly timestep mode, there's no progress output (eg telling the timestep).  I was going to fix that and send an update.

build/source/netcdf/def_output.f90
andywood: thanks for looking these over

build/source/netcdf/def_output.f90
martynpclark: You also do not need the second `//`

build/source/netcdf/def_output.f90
martynpclark: OK, that would be good. Would you want to add a separate PR for this so (as you say) it does not get lost?

build/source/engine/computFlux.f90
martynpclark: But it is adding to the earlier assignment.... it doesn't look right to me.

build/source/netcdf/def_output.f90
andywood: right!

build/source/engine/updateVars.f90
martynpclark: OK.

build/source/netcdf/def_output.f90
andywood: sure, I can separate them.

build/source/engine/computFlux.f90
andywood: I think it is overwriting the earlier assignment, not adding to it.

martynpclark: I really don't see this. `scalarRunoff` is calculated in soil hydrology (adding drainage) and then calculated again here (adding to it). I am probably missing something really obvious. This still does not seem right to me.

andywood:
This latest PR takes out the netcdf type change - will do it separately.  It responds to changes requested by MC in the earlier PR.  It also includes some additional minor changes related to time-checking for actions like progress reports.

build/source/engine/computFlux.f90
andywood: The second statement doesn't add it to, but replaces its value, so that when you leave computFlux you end up with the value including the aquifer.  Stepping through it in gdb looks like this:  
(gdb) s
797        scalarTotalRunoff  = scalarSurfaceRunoff + scalarAquiferBaseflow
(gdb) p scalartotalrunoff 
$11 = 1.2939818142135093e-08
(gdb) p scalarSurfaceRunoff
$12 = 0
(gdb) p scalarAquiferBaseflow
$13 = 0.00022222222222222215
(gdb) s
814      if(ixCasNrg/=integerMissing) fluxVec(ixCasNrg) = scalarCanairNetNrgFlux/canopyDepth
(gdb) p scalarTotalRunoff
$14 = 0.00022222222222222215
(gdb) 
-----
I think if it were:  scalarTotalRunoff  = scalarTotalRunoff + scalarSurfaceRunoff + scalarAquiferBaseflow
then it would be adding to the earlier value.

andywood: The second statement doesn't add it to, but replaces its value, so that when you leave computFlux you end up with the value including the aquifer.  Stepping through it in gdb looks like this:  
(gdb) s
797        scalarTotalRunoff  = scalarSurfaceRunoff + scalarAquiferBaseflow
(gdb) p scalartotalrunoff 
$11 = 1.2939818142135093e-08
(gdb) p scalarSurfaceRunoff
$12 = 0
(gdb) p scalarAquiferBaseflow
$13 = 0.00022222222222222215
(gdb) s
814      if(ixCasNrg/=integerMissing) fluxVec(ixCasNrg) = scalarCanairNetNrgFlux/canopyDepth
(gdb) p scalarTotalRunoff
$14 = 0.00022222222222222215
(gdb) 
-----
I think if it were:  scalarTotalRunoff  = scalarTotalRunoff + scalarSurfaceRunoff + scalarAquiferBaseflow
then it would be adding to the earlier value.

andywood: Sorry, I don't know why it's going all bold here

andywood: I could also be overlooking something as you certainly do more coding than I do.

martynpclark: Yes, I was missing something obvious. I read scalarSurfaceRunoff as scalarTotalRunoff.

andywood:
noticed one error in the new progress frequency option: comment line should note that  '_1d' is the default, not '_1t'.  I'll fix it in the next PR.

PR #377, opened 2020-03-25
1) write hruId to all output files (using same function with hru)

andywood:
2) avoid opening and default-writing timestep variables
3) remove pre-loaded outputfile suffixes
4) miscellaneous typo fixes

### DIFF
--- a/build/source/driver/summa_defineOutput.f90
+++ b/build/source/driver/summa_defineOutput.f90
@@ -18,12 +18,11 @@
 ! You should have received a copy of the GNU General Public License
 ! along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-module summa_defineOutput
-! used to define model output files
+module summa_defineOutput                     ! used to define model output files
 
 ! access missing values
-USE globalData,only:integerMissing   ! missing integer
-USE globalData,only:realMissing      ! missing double precision number
+USE globalData,only:integerMissing            ! missing integer
+USE globalData,only:realMissing               ! missing double precision number
 
 ! named variables to define new output files
 USE globalData, only: noNewFiles              ! no new output files
@@ -32,7 +31,7 @@ USE globalData, only: newFileEveryOct1        ! create a new file on Oct 1 every
 ! metadata structures
 USE globalData,only:attr_meta                 ! attributes metadata structure
 USE globalData,only:type_meta                 ! veg/soil type metadata structure
-USE globalData,only:id_meta                   ! veg/soil type metadata structure
+USE globalData,only:id_meta                   ! hru and gru Id metadata structure
 USE globalData,only:mpar_meta                 ! local parameter metadata structure
 USE globalData,only:bpar_meta                 ! basin parameter metadata structure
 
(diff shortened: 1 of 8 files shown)


andywood:
please take a look & think about whether these changes are desired & done in the best way.  I think having hruId in the output files makes a lot of sense when we run split domains and need to merge files; also not writing timestep files unless they're needed will make the model lighter for some apps.  Not forcing the user to have 'output' or 'spinup' in output file names is a matter of preference.  Typically all my output paths already have the word 'output' somewhere so I tend to leave it out of the actual output filename if I can ... a VIC-style user choice.  I'm not sure what the 'spinup' means wrt water year files but that's also now a choice that can be put in the outputControl file.  If these changes are acceptable there are a few compilation warnings that could be cleaned up; and I left commented out lines in the code for ease of re-adding them.  They'd be taken out and I'd generate a new clean PR.

build/source/driver/summa_defineOutput.f90
```diff
@@ -101,11 +100,13 @@ subroutine summa_defineOutputFiles(modelTimeStep, summa1_struc, err, message)
...
  if(modelTimeStep==1)then
   select case(newOutputFile)
-   case(noNewFiles);       fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//'output'//trim(output_fileSuffix)
-   case(newFileEveryOct1); fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//'spinup'//trim(output_fileSuffix)
+   !case(noNewFiles);       fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//'output'//trim(output_fileSuffix)
+   case(noNewFiles);       fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//trim(output_fileSuffix)
+   !case(newFileEveryOct1); fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//'spinup'//trim(output_fileSuffix)
+   case(newFileEveryOct1); fileout = trim(OUTPUT_PATH)//trim(OUTPUT_PREFIX)//trim(output_fileSuffix)
    case default; err=20; message=trim(message)//'unable to identify the option to define new output files'; return
   end select
```
martynpclark: It would be cleaner to remove commented lines

build/source/driver/summa_defineOutput.f90
```diff
@@ -133,7 +134,7 @@ subroutine summa_defineOutputFiles(modelTimeStep, summa1_struc, err, message)
     select case(trim(structInfo(iStruct)%structName))
      case('attr'); call writeParm(gru_struc(iGRU)%hruInfo(iHRU)%hru_ix,attrStruct%gru(iGRU)%hru(iHRU),attr_meta,err,cmessage)
      case('type'); call writeParm(gru_struc(iGRU)%hruInfo(iHRU)%hru_ix,typeStruct%gru(iGRU)%hru(iHRU),type_meta,err,cmessage)
-     case('id');   call writeParm(gru_struc(iGRU)%hruInfo(iHRU)%hru_ix,  idStruct%gru(iGRU)%hru(iHRU),  id_meta,err,cmessage)
+     !case('id');   call writeParm(gru_struc(iGRU)%hruInfo(iHRU)%hru_ix,  idStruct%gru(iGRU)%hru(iHRU),  id_meta,err,cmessage)
      case('mpar'); call writeParm(gru_struc(iGRU)%hruInfo(iHRU)%hru_ix,mparStruct%gru(iGRU)%hru(iHRU),mpar_meta,err,cmessage)
     end select
     if(err/=0)then; message=trim(message)//trim(cmessage)//'['//trim(structInfo(iStruct)%structName)//']'; return; endif
```
martynpclark: Why is this commented?

build/source/dshare/popMetadat.f90
```diff
@@ -781,7 +780,7 @@ subroutine read_output_file(err,message)
  outFreq(:) = .false.
```
martynpclark: Why is ID removed here?

martynpclark:
The code looks good to me. Once you get comfortable with what you've done it would be good to remove some of the lines that are commented out.

Have you tested this yet?

build/source/driver/summa_defineOutput.f90
andywood: right, thanks - I left them in in case you advised not to make the changes, in which case it would be easier for me to revert them.  I was planning to remove them in a clean PR after making suggested changes.  but I can see this is a little messy.

build/source/driver/summa_defineOutput.f90
andywood: hruId is now written in the same function that puts hru in all the files, not by writeParm.  The advantage is that you don't have to specify hruId in your outputcontrol file along with other variables, & for every timestep requested.

build/source/dshare/popMetadat.f90
andywood: Id was being put into the timestep file and among other variables forcing it to be opened and written by default.

andywood:
I've been running this code and it works fine.  I did test in addition to writing timestep-level files that were the defaults before (eg with nSnow, nSoil etc), and it worked.

build/source/dshare/popMetadat.f90
andywood: It's now part of id_meta, as it's not categorical/'type' data).  id_meta should be expanded later to have gru_ids and hru2gru ids and possibly more, but I'll look into that in another effort.

PR #380, opened 2020-03-31
Updated docs with netcdf dimension note

wknoben:
Added a note to the docs (SUMMA Input - Meteorological Forcing Files) that different programming languages interact with NetCDF dimensions in different ways. Because SUMMA expects the forcing files to have dimension (hru,time) this could lead to problems for users using e.g. Python to generate their forcing files.


wknoben:
Resolved the merge conflict by keeping the more recent version of the develop branch docs intact.

PR #382, opened 2020-04-03
Improve output configuration

arbennett:
This pull request adds 2 new features. The first, is the ability to add `outPrecision` to the output control file to specify the output precision as either single or double precision. This can be set as:
```
outputPrecision | single
outputPrecision | float
outputPrecision | double
```

where `single` and `float` both are single-precision floating point numbers and `double` is double precision floating point numbers.

Additionally this defines the chunking for output variables in an explicit way. This helps to speed up writeout for smaller simulations as well as maintains a reasonable file size for larger simulations. As far as my testing went, setting the maximum chunksize for any variable type to ~1000 worked well, but this might be explored further in the future.


andywood:
Nice work, Andrew.  Having also explored both of these factor (float v double output and chunksizes), I think it's great to make these changes.  That said, it would be good to get some more info/benchmarking on the choices for settings (eg 480, 960 etc) for how things are chunked in this commit.  For the small runs the increase in time chunksize makes a huge difference, but I've also seen it be problematic in file sizes for really large hru-dim runs.  

Aside from this I'd suggest making the the time & hru/gru global chunksizes into params that could be set in the 'localParamInfo' (ie 'globalHruParamInfo') in the same section as the balance convergence tolerances.   To optimize for a given application (eg small scale runs vs large domain runs), a user could just change those settings.   

Also, btw, if 'single' and 'float' lead to the same outcome, why not just make it 'single' versus' 'double' as options?  or 'float' versus 'double'?

arbennett:
I think it's probably worth doing a bit of benchmarking on a really large run before this is merged, and think we could do some balancing if the number of HRU or GRU is > 1000. In that case we might do something like equally balance the `timeChunk` and `gruChunk`/`hruChunk` so that both have the same number of chunks. That way both timeseries and spatial operations are somewhat balanced in terms of processing time.

build/source/dshare/popMetadat.f90
```diff
@@ -791,6 +792,16 @@ subroutine read_output_file(err,message)
...
   ! user cannot control time output
   if (trim(varName)=='time') cycle
+  ! set precision if it is given
+  if (trim(varName)=='outputPrecision') then
+    statName = trim(lineWords(nWords))
+    if (statName=='single' .or. statName=='float') then
+      outputPrecision = nf90_float
+    else if (statName=='double') then
+      outputPrecision = nf90_double
+    end if
```
bartnijssen, line 802: error out if neither

build/source/netcdf/def_output.f90
```diff
@@ -149,18 +151,18 @@ subroutine def_output(summaVersion,buildTime,gitBranch,gitHash,nGRU,nHRU,nSoil,i
...
-    case('diag' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,diag_meta, nf90_double,err,cmessage)  ! model diagnostic variables
-    case('flux' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,flux_meta, nf90_double,err,cmessage)  ! model fluxes
-    case('bvar' ); call def_variab(ncid(iFreq),iFreq,needGRU,needTime,bvar_meta, nf90_double,err,cmessage)  ! basin-average variables
+    case('forc' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,forc_meta, outputPrecision, err,cmessage)  ! model forcing data
+    case('prog' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,prog_meta, outputPrecision, err,cmessage)  ! model prognostics
+    case('diag' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,diag_meta, outputPrecision, err,cmessage)  ! model diagnostic variables
+    case('flux' ); call def_variab(ncid(iFreq),iFreq,needHRU,needTime,flux_meta, outputPrecision, err,cmessage)  ! model fluxes
+    case('bvar' ); call def_variab(ncid(iFreq),iFreq,needGRU,needTime,bvar_meta, outputPrecision, err,cmessage)  ! basin-average variables
```
bartnijssen, line 165: Align comments

build/source/netcdf/def_output.f90
```diff
@@ -284,7 +288,12 @@ subroutine def_variab(ncid,iFreq,spatialDesire,timeDesire,metaData,ivtype,err,me
  integer(i4b)                  :: iVarId            ! netcdf variable index
  integer(i4b)                  :: iStat             ! stat index
  integer(i4b),allocatable      :: dimensionIDs(:)   ! vector of dimension IDs
+ integer(i4b),allocatable      :: writechunk(:)      ! size of chunks to be written
```
bartnijssen, line 291: align comment

build/source/netcdf/modelwrite.f90
```diff
@@ -552,14 +552,14 @@ subroutine writeRestart(filename,         & ! intent(in): name of restart file
...
+   case(iLookvarType%wLength);                err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,  specDimID /),ncVarID(iVar))
+   case(iLookvarType%midSoil);                err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,midSoilDimID/),ncVarID(iVar))
+   case(iLookvarType%midToto);                err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,midTotoDimID/),ncVarID(iVar))
+   case(iLookvarType%ifcSoil);                err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,ifcSoilDimID/),ncVarID(iVar))
+   case(iLookvarType%ifcToto);                err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,ifcTotoDimID/),ncVarID(iVar))
+   case(iLookvarType%midSnow); if (maxSnow>0) err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,midSnowDimID/),ncVarID(iVar))
+   case(iLookvarType%ifcSnow); if (maxSnow>0) err = nf90_def_var(ncid,trim(prog_meta(iVar)%varname),nf90_float,(/hruDimID,ifcSnowDimID/),ncVarID(iVar))
```
bartnijssen, line 562: Revert to `nf90_double` for restart file - should not be user-selectable to enforce exact restarts

docs/input_output/SUMMA_input.md
```diff
@@ -139,6 +139,8 @@ scalarSenHeatTotal | 24      | 0   | 1    | 1    | 0   | 1   | 1   | 0
 ```
 In this example, the first line is a comment (starts with `!`) and then the sum, mean, min, max are calculated for `scalarSenHeatTotal` across 24 forcing time steps and written to the output file.
 
+Additionally, you can specify the output precision as `float`, `single`, or `double` by adding the line `outputPrecision | single` to the output control file. The default precision if this is not included is `double`. Both `single` and `float` correspond to single precision.
```
bartnijssen, line 142: Fix logic

PR #385, opened 2020-04-20
Add ability to have minutes as reference timestep

arbennett:
This just allows `minutes` to be the reference time units from the given reference time for forcing data.


andywood:
Funny, I just made the same change ... good to include this.

PR #386, opened 2020-04-23
Add WARNING when file manager input not found.  

andywood:
This can lead to run errors that are difficult to decipher if a user is not expecting that other default parameters are being used.  The warning appears as below (followed by an example of the errors occurring subsequently, which don't indicate that the user's main problem is a path error leading to skipping the reading of the trialParams file).  

...
  27 thCondSnow: jrdn1991
  28 thCondSoil: funcSoilWet
  29 spatial_gw: localColumn
  30 subRouting: timeDlay
startTime: iyyy, im, id, ih, imin = 1990  8  1  2  0
finshTime: iyyy, im, id, ih, imin = 1991 10  2 23  0
number of time steps =        3424
Skipping over LUTYPE = USGS

**!! WARNING:  trial parameter file not found; proceeding instead with other default parameters; check path in file manager input if this was not the desired behavior**

 theta_res         =    0.0000000000000000        0.0000000000000000        0.0000000000000000     
 theta_sat         =    0.0000000000000000        0.0000000000000000        0.0000000000000000     
 critSoilTranspire =   0.35999999999999999     

FATAL ERROR: summa_paramSetup/paramCheck/critSoilTranspire parameter is out of range [NOTE: if overwriting Noah-MP soil table values in paramTrial, must overwrite all soil parameters]


andywood:
There is some code in commit a377ad9 that is not yet needed.  The idea was to store all the filemanager options in a structure similar to the decisions options, but most of those modules needed are not available by the time the filemanager code is compiled -- it's one of the first.  The code for those structures is still there but would be deleted if there's no way or desire to do this.

andywood:
The latest commit in this PR just adds gru/gruId to the output files.  I didn't have time to split hru/gru output files as discussed, but this will remove the step of creating/adding gruId to basin runoff before it can be routed by mizuRoute.  At this point the summa output files can be routed directly (or after merging from a split domain run)

build/source/driver/summa_alarms.f90
```diff
@@ -86,7 +86,7 @@ subroutine summa_setWriteAlarms(oldTime, newTime, endTime,       &   ! time vect
  character(*),intent(out)              :: message            ! error message
  ! ---------------------------------------------------------------------------------------
  ! local variables
- character(LEN=256)                    :: cmessage           ! error message of downwind routine
+ !character(LEN=256)                    :: cmessage           ! error message of downwind routine
```
bartnijssen, line 89: never used - delete

bartnijssen:
Closing -- will be replaced by a number of small PRs

build/source/driver/summa_alarms.f90
andywood: Not really 'never used' ... rather it's good to see this PR's restructuring of summa control files and other usability changes (eg writing GRU ids) merged into the master via your subsequent split PRs (leading in part to version 3.0).  

It looks like there may be one discrepancy with this original that's now an error the current code -- nf90_def_var(ncid, 'gruId', nf90_int64, gru_DimID, gruIdVarID); -- is now writing 'nf90_int' -- I'm not sure where this change happened, since the newer version worked fine up to now (v3.0.2).  Trying to track it ...

andywood: hopefully this is not me looking at a file from the wrong branch

PR #388, opened 2020-05-13
add instructions to the Makefile

martynpclark:


build/Makefile
```diff
@@ -26,6 +26,8 @@
 # following unless you are doing SUMMA development and changed what 
 # needs to be compiled
 
+# Have fun
+
 #========================================================================
```
martynpclark, line 31: Do you really need to be so prescriptive in your instructions

martynpclark: Yes, please merge. It is important that people have fun.

PR #391, opened 2020-06-14
Fix writechunk declaration if statements

arbennett:
This has passed the testcases bit for bit.


andywood:
Bug found/fixed by Hongli Liu

PR #392, opened 2020-06-18
Fix build

arbennett:


bartnijssen:
Build is passing after recent updates to the Makefile - fix no longer needed

PR #407, opened 2020-07-18
Initial commit of v2 to v3 configuration converter

bartnijssen:
This will update the v2 configuration files to v3 - can be improved


andywood:
nice addition

PR #408, opened 2020-07-19
Rename variables for file manager, update conversion script

arbennett:


arbennett:
Mainly I don't like `outputDefFile` for aesthetic reasons, and typo-ability because of the `fF` portion. I'm fine with some other name as well, perhaps `outputDefnFile`? I think `spatialParams` is a clear improvement over `trialParams` - that has some implication that you'd try these and if they don't work do something else. However, this file _is_ how you include spatially distributed parameters, hence `spatialParams`. The tables are an entirely separate problem to me, and I really don't like that they are implicit in SUMMA setups, but that's for another time. 

Just one other thought, I personally think it's pretty obvious that these should be files, given the keys don't say "path" anywhere. I would be more in favor of dropping the`File` from all of these in order to make space for better naming. Thoughts on this?

PR #412, opened 2020-07-28
Adding optional statePath entry to file manager to facilitate file ma…

andywood:
…nagement, eg hindcasting. This was overlooked in my recent reconfiguration of the fileManager for the same purpose.  This is backward compatible.

### DIFF
--- a/build/source/driver/summa_init.f90
+++ b/build/source/driver/summa_init.f90
@@ -44,7 +44,8 @@ USE globalData,only:statIndx_meta                           ! child metadata for
 USE globalData,only:statBvar_meta                           ! child metadata for stats
 
 ! provide access to file paths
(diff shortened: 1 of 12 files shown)


andywood:
this adds needed functionality to streamline a hindcast workflow by allowing output state files to be used as input state files from the same place, and without forcing a mix of state & output files in the output file directory.  It's backward compatible so it can be ignored and the current default behavior will continue.  

I also renamed INPUT_PATH to FORCING_PATH to be more specific (since settings are also model inputs), and expanded SETNGS_PATH to SETTINGS_PATH for readability, since there doesn't at this point seem to be any reason to adhere to some shorter variable name length.  

I spot checked this using a run with and without using statePath in the fileManager.

PR #420, opened 2020-09-08
fix hru2gru mapping

martynpclark:
The mapping between HRUs and GRUs in the `index_map` structure was not correct. Fixed.


bartnijssen:
@arbennett : Can you review this and merge (or let me know to merge)? The PR is against `master`, which is probably OK, but then make sure to merge it back into `develop`

arbennett (approved):

PR #426, opened 2020-09-16
avoid very small stability values

martynpclark:
The "standard" option for the stability correction is zero above the critical Richardson number. This provides an angle discontinuity -- close to the critical Richardson number values. The other stability corrections are more smooth.

To reduce numerical problems, the stability correction for the "standard" option sets stability to epsilon (~10^-16) if the bulk Richardson number is higher than the critical Richardson number. This still causes convergence problems. Suggest a minor modification to set the stability correction to 10^-10 if the Richardson number is above the critical value.

### DIFF
--- a/build/source/engine/vegNrgFlux.f90
+++ b/build/source/engine/vegNrgFlux.f90
@@ -3140,6 +3140,7 @@ contains
  integer(i4b),intent(out)      :: err                    ! error code
  character(*),intent(out)      :: message                ! error message
  ! local
+ real(dp), parameter           :: verySmall=1.e-10_dp    ! a very small number (avoid stability of zero)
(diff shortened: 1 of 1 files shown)


bartnijssen (commented):
Wouldn't it make sense to simply update the second occurrence to `epsilon(stabilityCorrection)`? Especially since that value may be something we could expose to the user.

Whoops - I guess I should read your actual description first!

martynpclark:
Just for completeness, `epsilon(stabilityCorrection)` is a small number (epsilon) of type stabilityCorrection, i.e., EPSILON(X) returns the smallest number E of the same kind as X such that 1 + E > 1.

bartnijssen:
Oh I see - I did not realize fortran had an intrinsic function for that purpose. Disregard - I'll approve and merge the PR.

PR #430, opened 2020-09-22
Fix restart file write/read to include basin hillslope runoff variabl…

andywood:
…es -- allowing for initialized predictions or simulations that need basin runoff.
State files for prior versions will no longer match this version, but also won't support exact restarts for basin variables.


andywood:
The PR #430 fix hardwires a variable (nTDH) in two places that could be passed from allocGlobal ... given pressing deadlines I have to leave that to someone with more time to adjust.  This has been tested on a 16-GRU basin.  Next test will include a split domain run.

arbennett (approved):

PR #432, opened 2020-09-24
Bugfix/restart split domain

arbennett:
This fixes a bug in using restart files if you use the `-g startGRU countGRU` flag. Previously the initial conditions were always selected starting at the first index instead of at `startGRU`. Note that this does break the ability to use subsetted restart files.

### DIFF
--- a/build/source/netcdf/read_icond.f90
+++ b/build/source/netcdf/read_icond.f90
@@ -276,7 +276,6 @@ contains
 
   ! get data
   err = nf90_get_var(ncID,ncVarID,varData); call netcdf_err(err,message)
-  write(*,*) shape(varData), shape(gru_struc), ixFile
   if(err/=0)then; message=trim(message)//': problem getting the data for variable '//trim(prog_meta(iVar)%varName); return; endif
 
   ! store data in prognostics structure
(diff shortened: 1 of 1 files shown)


build/source/netcdf/read_icond.f90
```diff
@@ -292,10 +294,9 @@ subroutine read_icond(iconFile,                      & ! intent(in):    name of
     ! get the index in the file: single HRU
     if(restartFileType==singleHRU)then
      ixFile = 1  ! use for single HRU restart file
-
     ! get the index in the file: multi HRU
     else
-     ixFile = iHRU_local
+     ixFile = startGRU + iHRU_local - 1
```
bartnijssen, line 299: Why the `-1` offset relative to `startGRU`? That seems a C things (if `startGRU` starts at ` and indices start at 0). Is this correct? I guess if both `startGRU` and `iHRU_local` are 1 this works.

bartnijssen (commented):
@arbennett : Just see the question and provide a response.

build/source/netcdf/read_icond.f90
andywood: right.  I think this is the equivalent of using ixFile = iHRU_global, incidentally.  using iHRU_local was the bug before.

andywood (approved):

build/source/netcdf/read_icond.f90
arbennett: Exactly, basically it's just accounting for 2 things that both start at 1

PR #433, opened 2020-09-24
Output gruid and hruid as coordinates of gru and hru

### LINKED ISSUE #393: Make hruId and gruId coordinates for hru and gru

arbennett:
This simply makes the `gru` and `hru` dimensions have associated coordinates that instead of being sequential are now IDs. Closes #393

### DIFF
--- a/build/source/netcdf/def_output.f90
+++ b/build/source/netcdf/def_output.f90
@@ -469,14 +469,14 @@ contains
  do iGRU = 1, size(gru_struc)
 
   ! GRU info
-  err = nf90_put_var(ncid, gruVarID, gru_struc(iGRU)%gru_nc, start=(/iGRU/))
+  err = nf90_put_var(ncid, gruVarID, gru_struc(iGRU)%gruId, start=(/iGRU/))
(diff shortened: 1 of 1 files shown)


build/source/netcdf/def_output.f90
```diff
@@ -469,14 +469,14 @@ subroutine write_hru_info(ncid, err, message)
  do iGRU = 1, size(gru_struc)
 
   ! GRU info
-  err = nf90_put_var(ncid, gruVarID, gru_struc(iGRU)%gru_nc, start=(/iGRU/))
+  err = nf90_put_var(ncid, gruVarID, gru_struc(iGRU)%gruId, start=(/iGRU/))
```
bartnijssen, line 472: Is it really gruID and hru_id? That seems somewhat inconsistent (but a separate issue as long as this is correct)

bartnijssen (changes requested):

build/source/netcdf/def_output.f90
arbennett: Yep :(

bartnijssen (approved):

PR #435, opened 2020-10-10
Bug fix/too much melt

martynpclark:
In rare cases the solution for snow temperature was greater then Tfreeze. If this happens we reduce the length of the time step and try again.


andywood (commented):
perhaps expand the error messaging.  Most of my time in dealing with summa errors involves figuring out where they occur in the code -- so when we come across new possible errors we need to make the code explicit about what's happening.  It will save some user time in the future.  in this case it's a trivial detail, it's more the principle:

  ! if water balance error or toomuchmelt error then reduce the step length
 if(waterBalanceError .or. tooMuchMelt)then
   message=trim(message)//'water balance error or tooMuchMelt error'
   reduceCoupledStep=.true.
err=-20; return

grepping 'tooMuchMelt error in the code will lead right here...

PR #436, opened 2020-10-16
Bug fix/hyd cond

martynpclark:
The code to compute hydraulic conductivity in macropores is
```
 if(volFracLiq > theta_mp)then
  theta_e       = (volFracLiq - theta_mp) / (theta_sat - theta_mp)
  hydCondMP_liq = (satHydCond_ma - satHydCond_mi) * (theta_e**mpExp)
 else
```
which means that the hydraulic conductivity for macropores will be negative if the conductivity for macropores is greater than for micropores.

The total hydraulic conductivity is
```
scalarHydCond   = hydCond_noIce*iceImpedeFac + scalarHydCondMP
```
and the total conductivity can be negative if there is substantial ice impedance.

It is a good idea to require that the conductivity for macropores be greater than for micropores.

### DIFF
--- a/build/source/engine/var_derive.f90
+++ b/build/source/engine/var_derive.f90
@@ -326,13 +326,13 @@ contains
    case(powerLaw_profile)
     ! - conductivity at layer interfaces
     !   --> NOTE: Do we need a weighted average based on layer depth for interior layers?
-    
+
     if(compactedDepth/iLayerHeight(nLayers) /= 1._dp) then    ! avoid divide by zero
      ifcDepthScaleFactor = ( (1._dp - iLayerHeight(iLayer)/iLayerHeight(nLayers))**(zScale_TOPMODEL - 1._dp) ) / &
                            ( (1._dp -       compactedDepth/iLayerHeight(nLayers))**(zScale_TOPMODEL - 1._dp) )
     else
      ifcDepthScaleFactor = 1.0_dp
-    endif                           
+    endif
     if(iLayer==nSnow)then
      iLayerSatHydCond(iLayer-nSnow) = k_soil(1) * ifcDepthScaleFactor
     else   ! if the mid-point of a layer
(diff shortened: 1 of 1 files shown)


arbennett:
Could we simply set the conductivity to zero in this case?

bartnijssen:
@arbennett : Which conductivity would be set to 0?

Wouldn't it be easier to set the `k_macropore` to `k_soil` if `k_macropore < k_soil` and issue a warning, just to make sure that the model continues. If we can modify the message to indicate what a user can do to fix this that may be helpful as well.

```
if(k_macropore(iLayer-nSnow) < k_soil(iLayer-nSnow))then
      k_macropore(iLayer-nSnow) = k_soil(iLayer-nSnow)
      message=trim(message)//"hydraulic conductivity for macropores is less than the hydraulic conductivity for micropores"
      err=20; return
     endif
```

martynpclark:
@bartnijssen yes, this is a good idea. A better solution. It means that the model does not stop.

arbennett:
Yeah, I agree, @bartnijssen that sounds like a good solution.

andywood:
Yeah, that was my suggestion earlier too ...

andywood:
I would also expand the message, to "hydraulic conductivity for macropores
(VALUE) is less than the hydraulic conductivity for micropores (VALUE):
resetting macropore conductivity to equal micropore value" and even give
the gru/hru id if that's handy in the routine.

arbennett (changes requested):
It seems @andywood, @bartnijssen, and myself all would like to see the workaround so that this doesn't stop the simulation. I think what we really want to ensure is that at line 1031 of `soilLiqFlxf.90`:
```
scalarHydCond = max(hydCond_noIce*iceImpedeFac + scalarHydCondMP, 0.0_dp) # or maybe some tiny value
```

martynpclark:
Implemented changes in a separate PR

PR #443, opened 2021-01-05
initial implementation of lookup tables for enthalpy

martynpclark:
This is the initial implementation of the lookup tables for enthalpy. It just builds the infrastructure to do the "real" work later. Most of the code is in `t2enthalpy` and the enthalpy calculations are in `eval8summa`. To activate the enthalpy code set `needEnthalpy=.true.` in `eval8summa`.

### DIFF
--- a/build/source/engine/t2enthalpy.f90
+++ b/build/source/engine/t2enthalpy.f90
@@ -86,14 +86,14 @@ contains
  character(*),intent(out)      :: message              ! error message
  ! declare local variables
  character(len=128)            :: cmessage             ! error message in downwind routine
- logical(lgt),parameter        :: doTest=.false.       ! flag to test
  integer(i4b),parameter        :: nLook=100            ! number of elements in the lookup table
  integer(i4b),parameter        :: nIntegr8=10000       ! number of points used in the numerical integration
  real(dp),parameter            :: T_lower=260.0_dp     ! lowest temperature value where all liquid water is assumed frozen (K)
  real(dp),dimension(nLook)     :: xTemp                ! temporary vector
  real(dp)                      :: xIncr                ! temporary increment
  real(dp)                      :: T_incr               ! temperature increment
- real(dp),parameter            :: T_test=272.9742_dp   ! test value for temperature (K)
+ logical(lgt),parameter        :: doTest=.false.       ! flag to test
+ real(dp),parameter            :: T_test=249.9742_dp   ! test value for temperature (K)
  real(dp)                      :: E_test               ! test value for enthalpy (J m-3)
  integer(i4b)                  :: iVar                 ! loop through variables
  integer(i4b)                  :: iSoil                ! loop through soil layers
--- a/build/source/engine/t2enthalpy.f90
+++ b/build/source/engine/t2enthalpy.f90
@@ -1,0 +1,468 @@
+! SUMMA - Structure for Unifying Multiple Modeling Alternatives
+! Copyright (C) 2014-2020 NCAR/RAL; University of Saskatchewan; University of Washington
+!
+! This file is part of SUMMA
+!
+! For more information see: http://www.ral.ucar.edu/projects/summa
+!
+! This program is free software: you can redistribute it and/or modify
+! it under the terms of the GNU General Public License as published by
+! the Free Software Foundation, either version 3 of the License, or
+! (at your option) any later version.
+!
+! This program is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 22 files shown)


build/source/engine/t2enthalpy.f90
```diff
@@ -0,0 +1,468 @@
...
+    Tk(iLook)  = Tk(iLook) + T_incr
+
+    ! compute the volumetric liquid water and ice content
+    ! NOTE: assume saturation
+    matricHead = (LH_fus/gravity)*(Tk(iLook) - Tfreeze)/Tfreeze
+    vFracLiq   = volFracLiq(matricHead,vGn_alpha,theta_res,theta_sat,vGn_n,vGn_m)
+    vFracIce   = theta_sat - vFracLiq
+
+    ! compute enthalpy
+    ! NOTE: assume intrrinsic density of ice is the intrinsic density of water
```
wknoben, line 199: small typo in "intrinsic"

build/source/engine/t2enthalpy.f90
```diff
@@ -0,0 +1,468 @@
...
+   T_incr = (xTemp(iLook)-xTemp(iLook+1))/real(nIntegr8, kind(dp))
+
+   ! numerical integration between different values of the lookup table
+   do jIntegr8=1,nIntegr8
+
+    ! update temperature
+    Tk(iLook)  = Tk(iLook) + T_incr
+
+    ! compute the volumetric liquid water and ice content
+    ! NOTE: assume saturation
```
wknoben, line 193: Is this a safe assumption? Are there reasons why we don't compute this on each time step with the current values for vFracLiq and vFraqIce?

build/source/engine/t2enthalpy.f90
```diff
@@ -0,0 +1,468 @@
...
+     ! *** compute enthalpy of water for unfrozen conditions
+     if(mlayerTempTrial(iLayer) > Tcrit)then
+      enthWater = iden_water*Cp_water*vFracWat*(mlayerTempTrial(iLayer) - Tfreeze) ! valid for temperatures below freezing also
+      enthPhase = 0._dp
+
+     ! *** compute enthalpy of water for frozen conditions
+     else
+
+      ! calculate enthalpy at the temperature (cubic spline interpolation)
+      call splint(Tk,Ey,E2,mlayerTempTrial(iLayer),enthTemp,err,cmessage)
```
wknoben, line 421: These tables are calculated assuming saturated soil. Does that assumption hold when we get to this point in the code?

build/source/engine/t2enthalpy.f90
```diff
@@ -0,0 +1,468 @@
...
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU General Public License for more details.
+!
+! You should have received a copy of the GNU General Public License
+! along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+module t2enthalpy_module
+
+! constants
+USE multiconst, only: gravity, &                          ! gravitational acceleration (m s-1)
```
wknoben, line 24: Typo in units, should be (m s-2).
(Otherwise the matric head calculation on line 194 doesn't work out)

build/source/engine/t2enthalpy.f90
```diff
@@ -0,0 +1,468 @@
...
+
+   ! numerical integration between different values of the lookup table
+   do jIntegr8=1,nIntegr8
+
+    ! update temperature
+    Tk(iLook)  = Tk(iLook) + T_incr
+
+    ! compute the volumetric liquid water and ice content
+    ! NOTE: assume saturation
+    matricHead = (LH_fus/gravity)*(Tk(iLook) - Tfreeze)/Tfreeze
```
wknoben, line 194: Might be helpful to include a reference for these equations. Is this re-written from Flerchinger & Saxton (1989) and Fuchs et al. (1978)? 

Flerchinger, G. N., and Saxton, K. E.: Simultaneous heat and water model o f a freezing snow-residue-620soil system: 1. Theory and development, Trans. ASAE, 32, 565–571, 1989

Fuchs, M., Campbell, G.S. and Papendick, R.I. (1978), An Analysis of Sensible and Latent Heat Flow in a Partially Frozen Unsaturated Soil. Soil Science Society of America Journal, 42: 379-385. https://doi.org/10.2136/sssaj1978.03615995004200030001x

wknoben (changes requested):
Could you add a brief test that shows that if the enthalpy module is not enabled, simulations remain unchanged from a SUMMA version that does not include the enthalpy code?

Could you add a brief description of the checks used to show that this module produces the expected values?

PR #445, opened 2021-01-11
Fix for segfault on reference to `mw_exp` when using JRDN snow layering

arbennett:
Occasionally, when using JRDN snow layering I have run into segfaults on reference to `mw_exp` (only when doing scalar solutions) in the line:
```
iLayerLiqFluxSnow(iLayer)      = k_snow*relSaturn**mw_exp
```

after some debugging it turns out this loss of access to `mw_exp` occurs directly following the `pack` statement:
```
ixLayerDesired = pack(ixLayerState, ixSnowOnlyHyd/=integerMissing)
```

I have not been able to determine the exact reason for this error, but have developed a workaround, which is provided in this PR.

Solves #445

### DIFF
--- a/build/source/engine/snowLiqFlx.f90
+++ b/build/source/engine/snowLiqFlx.f90
@@ -92,8 +92,8 @@ contains
  character(*),intent(out)        :: message                    ! error message
  ! ------------------------------------------------------------------------------------------------------------------------------------------
  ! local variables
+ integer(i4b)                    :: i                          ! search index for scalar solution
  integer(i4b)                    :: iLayer                     ! layer index
- integer(i4b)                    :: ixLayerDesired(1)          ! layer desired (scalar solution)
  integer(i4b)                    :: ixTop                      ! top layer in subroutine call
  integer(i4b)                    :: ixBot                      ! bottom layer in subroutine call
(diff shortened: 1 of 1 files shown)


build/source/engine/snowLiqFlx.f90
```diff
@@ -131,10 +131,27 @@ subroutine snowLiqFlx(&
...
-  ixTop = ixLayerDesired(1)
-  ixBot = ixLayerDesired(1)
+  ! WARNING: Previously this was implemented as:
+  !    ixLayerDesired = pack(ixLayerState, ixSnowOnlyHyd/=integerMissing)
+  !    ixTop = ixLayerDesired(1)
+  !    ixBot = ixLayerDesired(1)
+  ! This implementation can result in a segfault when using JRDN layering.
+  ! The segfault occurs when trying to access `mw_exp` in:
+  !    iLayerLiqFluxSnow(iLayer)      = k_snow*relSaturn**mw_exp
+  ! Debugging found that the `pack` statement cased `mw_exp` to no longer be accessible.
```
bartnijssen, line 143: fix type 'cased' --> 'caused'

arbennett: Updated

PR #454, opened 2021-04-04
Addition of functionality for inputting aspect and non-zero slope/aspect effects on radiation, and also a BUG FIX: correcting averageRoutedRunoff in run_oneGRU.f90 and qTimeDelay.f90

andywood:


build/source/engine/read_attrb.f90
```diff
@@ -349,6 +349,20 @@ subroutine read_attrb(attrFile,nGRU,attrStruct,typeStruct,idStruct,err,message)
...
+ 
+ ! ** now handle the optional aspect variable if it's missing
+ varIndx = get_ixAttr('aspect')
+ ! check that the variable was not found in the attribute file
+ if(.not. checkAttr(varIndx)) then
+   write(*,*) NEW_LINE('A')//'INFO: aspect not found in the input attribute file, continuing ...'//NEW_LINE('A')
+
+   do iGRU=1,nGRU
+    do iHRU = 1, gru_struc(iGRU)%hruCount
+     attrStruct%gru(iGRU)%hru(iHRU)%var(varIndx) = -1._dp      ! populate variable with out-of-range value, used later
```
arbennett, line 361: Could this be changed to a missing value placeholder from `engine/nrtype.f90` so that the intention is clearer?

build/source/engine/derivforce.f90
```diff
@@ -234,11 +236,21 @@ subroutine derivforce(time_data,forc_data,attr_data,mpar_data,prog_data,diag_dat
  dataStep = data_step/secprhour  ! time step (hours)
  ahour    = real(jh,kind(dp)) + real(jmin,kind(dp))/minprhour - data_step/secprhour  ! decimal hour (start of the step)
 
+ ! check slope/aspect intent for radiation calculation
+ if(aspect == -1)then
```
arbennett, line 240: Could this be changed to a missing value indicator from `engine/nrtype.f90` to be clearer?

build/source/engine/computFlux.f90
```diff
@@ -793,9 +795,10 @@ subroutine computFlux(&
                    err,cmessage)                   ! intent(out): error control
    if(err/=0)then; message=trim(message)//trim(cmessage); return; endif
 
-   ! compute total runoff (overwrite previously calculated value before considering aquifer)
+   ! compute total runoff (overwrite previously calculated value before considering aquifer).  
+   !   (Note:  SoilDrainage goes into aquifer, not runoff)
    scalarTotalRunoff  = scalarSurfaceRunoff + scalarAquiferBaseflow
-
+
```
arbennett, line 801: Remove empty spaces.

build/source/engine/computFlux.f90
```diff
@@ -771,7 +773,7 @@ subroutine computFlux(&
 
  ! check if computing aquifer fluxes
  if(ixAqWat/=integerMissing)then
-
+
```
arbennett, line 776: Remove empty spaces.

build/source/engine/computFlux.f90
```diff
@@ -729,7 +729,7 @@ subroutine computFlux(&
     message=trim(message)//'expect dBaseflow_dMatric to be nSoil x nSoil'
     err=20; return
    endif
-
+
```
arbennett, line 732: Remove empty spaces.

build/source/engine/checkStruc.f90
```diff
@@ -152,10 +152,10 @@ subroutine checkPopulated(iStruct,metadata,err,message)
 
   ! loop through variables
   do iVar=1,size(metadata)
-
+
```
arbennett, line 155: Minor cleanup, but could you remove empty line spaces?

arbennett (changes requested):
Just a couple of minor style changes. Could you also add an entry to the `docs/whats-new.md`?

andywood:
Made all suggested changes

arbennett (approved):

build/source/engine/run_oneGRU.f90
```diff
@@ -219,29 +224,35 @@ subroutine run_oneGRU(&
...
+  ! otherwise just increment basin (GRU) column outflow (m3 s-1) with the hru fraction
   else
-   bvarData%var(iLookBVAR%basin__ColumnOutflow)%dat(1)   = bvarData%var(iLookBVAR%basin__ColumnOutflow)%dat(1) + sum(fluxHRU%hru(iHRU)%var(iLookFLUX%mLayerColumnOutflow)%dat(:))
+   bvarData%var(iLookBVAR%basin__ColumnOutflow)%dat(1) = bvarData%var(iLookBVAR%basin__ColumnOutflow)%dat(1) + sum(fluxHRU%hru(iHRU)%var(iLookFLUX%mLayerColumnOutflow)%dat(:))
   end if
 
+  ! ----- calculate weighted basin (GRU) fluxes --------------------------------------------------------------------------------------
+  
+  ! increment basin total runoff (m s-1)
+  bvarData%var(iLookBVAR%basin__TotalRunoff)%dat(1)    = bvarData%var(iLookBVAR%basin__TotalRunoff)%dat(1)     + fluxHRU%hru(iHRU)%var(iLookFLUX%scalarTotalRunoff)%dat(1)    * fracHRU
```
martynpclark, line 239: This variable will only work when the HRUs within a GRU are disconnected

build/source/engine/run_oneGRU.f90
```diff
@@ -219,29 +224,35 @@ subroutine run_oneGRU(&
...
 
   ! increment basin-average baseflow input variables (m s-1)
   bvarData%var(iLookBVAR%basin__AquiferRecharge)%dat(1)  = bvarData%var(iLookBVAR%basin__AquiferRecharge)%dat(1)   + fluxHRU%hru(iHRU)%var(iLookFLUX%scalarSoilDrainage)%dat(1)     * fracHRU
   bvarData%var(iLookBVAR%basin__AquiferTranspire)%dat(1) = bvarData%var(iLookBVAR%basin__AquiferTranspire)%dat(1)  + fluxHRU%hru(iHRU)%var(iLookFLUX%scalarAquiferTranspire)%dat(1) * fracHRU
 
-  ! increment aquifer baseflow -- ONLY if baseflow is computed individually for each HRU
+  ! increment aquifer baseflow -- ONLY if aquifer baseflow is computed individually for each HRU and aquifer is run
   ! NOTE: groundwater computed later for singleBasin
-  if(model_decisions(iLookDECISIONS%spatial_gw)%iDecision == localColumn)then
+  if(model_decisions(iLookDECISIONS%spatial_gw)%iDecision == localColumn .and. model_decisions(iLookDECISIONS%groundwatr)%iDecision == bigBucket) then
```
martynpclark, line 250: Do we need the additional condition that the `bigBucket` must be run? `aquiferBaseflow` is set to zero if there is no aquifer. Keeping the original code ensures that `basin__AquiferBaseflow` is also set to zero.

build/source/engine/run_oneGRU.f90
```diff
@@ -219,29 +224,35 @@ subroutine run_oneGRU(&
...
    bvarData%var(iLookBVAR%basin__AquiferBaseflow)%dat(1)  =  bvarData%var(iLookBVAR%basin__AquiferBaseflow)%dat(1)  &
-           +  fluxHRU%hru(iHRU)%var(iLookFLUX%scalarAquiferBaseflow)%dat(1) * fracHRU  &
-           +  fluxHRU%hru(iHRU)%var(iLookFLUX%scalarSoilDrainage)%dat(1)    * fracHRU
+           +  fluxHRU%hru(iHRU)%var(iLookFLUX%scalarAquiferBaseflow)%dat(1) * fracHRU
```
martynpclark, line 252: Yes, this is the bug. The fix looks correct.

build/source/engine/run_oneGRU.f90
```diff
@@ -260,9 +271,7 @@ subroutine run_oneGRU(&
  call qOverland(&
                 ! input
                 model_decisions(iLookDECISIONS%subRouting)%iDecision,          &  ! intent(in): index for routing method
-                bvarData%var(iLookBVAR%basin__SurfaceRunoff)%dat(1),           &  ! intent(in): surface runoff (m s-1)
-                bvarData%var(iLookBVAR%basin__ColumnOutflow)%dat(1)/totalArea, &  ! intent(in): outflow from all "outlet" HRUs (those with no downstream HRU)
-                bvarData%var(iLookBVAR%basin__AquiferBaseflow)%dat(1),         &  ! intent(in): baseflow from the aquifer (m s-1)
```
martynpclark, line 265: We do need these three arguments.

build/source/engine/qTimeDelay.f90
```diff
@@ -68,8 +64,8 @@ subroutine qOverland(&
  ! initialize error control
  err=0; message='qOverland/'
 
- ! compute instantaneous runoff (m s-1)
- averageInstantRunoff = averageSurfaceRunoff + averageAquiferBaseflow + averageSoilBaseflow
```
martynpclark, line 72: The original code was correct.

martynpclark (changes requested):

build/source/engine/read_attrb.f90
andywood: done

build/source/engine/derivforce.f90
andywood: done

build/source/engine/checkStruc.f90
andywood: done

build/source/engine/derivforce.f90
```diff
@@ -234,11 +236,21 @@ subroutine derivforce(time_data,forc_data,attr_data,mpar_data,prog_data,diag_dat
  dataStep = data_step/secprhour  ! time step (hours)
  ahour    = real(jh,kind(dp)) + real(jmin,kind(dp))/minprhour - data_step/secprhour  ! decimal hour (start of the step)
 
+ ! check slope/aspect intent for radiation calculation
+ if(aspect == nr_realMissing)then
+  azimuth = 0._dp              ! if aspect is not an input attribute, slope & azimuth = zero (flat Earth)
+  slope   = 0._dp
+ else
+  azimuth = aspect                               ! in degrees
+  slope   = atan(abs(tan_slope))*180.0D0/PI_D    ! convert from m/m to degrees
```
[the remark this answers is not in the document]
andywood: done

build/source/engine/run_oneGRU.f90
andywood: I don't see this ... could you explain it more?   It should be just like the terms below ...

build/source/engine/run_oneGRU.f90
andywood: I'm not sure - I was thinking we don't need to calculate it if it's zero (ie not run), and it's initialized to zero higher up (before the hru loop).

perhaps the two other aquifer summation variables above the if condition should also be included in this block.

build/source/engine/run_oneGRU.f90
andywood: can you elaborate?  these were removed to avoid calculating total runoff in 2 places ... the terms weren't used individually in qOverland.  I think the disconnect would be if there are more than one total runoff terms we want to calculate, eg on to use with the time delay, and one to use for something else.

build/source/engine/qTimeDelay.f90
andywood: agree ... but also duplicative of the calculation in run_oneGRU.  better to keep it in one place?

build/source/engine/run_oneGRU.f90
martynpclark: `scalarRunoff` is the runoff from an HRU. If HRUs are connected, then we only need the lateral flow from the most downslope HRU(s).

build/source/engine/run_oneGRU.f90
martynpclark: Yes, we could go either way I suppose. It could be easier to maintain if we reduce the number of conditions.

build/source/engine/run_oneGRU.f90
martynpclark: Basin runoff is not simply the sum of the HRU runoff in the case when there is a lateral flux across HRUs.

build/source/engine/qTimeDelay.f90
martynpclark: See comment above

martynpclark: I agree that it is best to remove the summation here.

build/source/engine/run_oneGRU.f90
martynpclark: I agree it is OK to add the if condition for `bigBucket`

build/source/engine/run_oneGRU.f90
martynpclark: We agreed to compute the total runoff in `run_oneGRU`. Hence these three arguments can be removed.

martynpclark (approved):

martynpclark:
The code looks good. Before merging, can I ask what tests were completed?

andywood:
apologies for not responding before - the main thing was to compare the new
basin runoff variable to total runoff and see that they matched.  I didn't
assess other variables.  The problem before was basin runoff being wrong by
a factor of about 2, which was 'fixed'.  I think better numerical tests
could be applied here, though.

PR #460, opened 2021-05-26
Bugfix/write gru parameters

martynpclark:


martynpclark:
This is just one line of additional code to enable writing GRU parameters. Previously there was a graceful failure if you added a GRU parameter to the ModelOutput file (e.g., a routing parameter). We now have a new case statement to accommodate the need to write GRU parameters.

PR #465, opened 2021-06-14
compute cpu time for each HRU and time step

martynpclark:

- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry


martynpclark:
This computes the CPU time spend on each HRU and time step. While the elapsed wall clock time is desired by users, the precision may be as low as 0.01 seconds. In many cases, it will be preferable to look at metrics like the number of flux calls. Nevertheless, the elapsed wall clock time is provided to meet user requests.

arbennett:
@martynpclark this looks good and I think is a nice feature to add, but I have one request - could you add a quick entry to the `docs/whats-new.md` list under a new header `Develop`?

martynpclark:
Updated whats-new file

PR #467, opened 2021-06-14
avoid graceful failure if using inappropriate routing parameters

martynpclark:

- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry


andywood (commented):
fine to merge - this improves on my earlier stab at an error message about this (which replaced an error message pointing to a non-existent routine/variable).

PR #470, opened 2021-06-17
revise Gamma distribution used for sub-grid routing

h294liu:

- [ ✓ ] closes #xxx (identify the issue associated with this PR)
- [ ✓ ] tests passed
- [ ✓ ] new tests added
- [ ✓ ] science test figures
- [ ✓ ] checked that the new code conforms to the [SUMMA coding conventions](https://github.com/NCAR/summa/blob/master/docs/howto/summa_coding_conventions.md)
- [ ] ReleaseNotes entry

The previous code has an error in calculating the cumulative probability "cumProb" of the Gamma distribution. Here the error is fixed by updating "sumProb" based on Equation 13 of Clark at el. (2007, WRR); removing variable "aLambda"; introducing variable "routingGammaMean".

The attached figures show the sensitivity of the updated Gamma distribution to the summa parameters routingGammaShape and routingGammaScale, respectively. 
shape_sensitivity
scale_sensitivity

### DIFF
--- a/build/source/engine/var_derive.f90
+++ b/build/source/engine/var_derive.f90
@@ -393,7 +393,7 @@ contains
  real(rkind)                        :: dt                     ! data time step (s)
  integer(i4b)                    :: nTDH                   ! number of points in the time-delay histogram
  integer(i4b)                    :: iFuture                ! index in time delay histogram
- real(rkind)                        :: aLambda                ! scale parameter in the Gamma distribution
+ real(rkind)                        :: routingGammaMean       ! mean of the Gamma distribution
  real(rkind)                        :: tFuture                ! future time (end of step)
  real(rkind)                        :: pSave                  ! cumulative probability at the start of the step
  real(rkind)                        :: cumProb                ! cumulative probability at the end of the step
@@ -433,8 +433,8 @@ contains
   case(timeDelay)
    ! initialize
    pSave   = 0._rkind ! cumulative probability at the start of the step
-   aLambda = routingGammaShape / routingGammaScale
-   if(routingGammaShape <= 0._rkind .or. aLambda < 0._rkind)then
+   routingGammaMean = routingGammaShape * routingGammaScale
+   if(routingGammaShape <= 0._rkind .or. routingGammaMean < 0._rkind)then
     message=trim(message)//'bad arguments for the Gamma distribution'
     err=20; return
    end if
@@ -442,7 +442,7 @@ contains
    do iFuture = 1,nTDH
     ! get weight for a given bin
     tFuture = real(iFuture, kind(dt))*dt                  ! future time (end of step)
-    cumProb = gammp(routingGammaShape,aLambda*tFuture)    ! cumulative probability at the end of the step
+    cumProb = gammp(routingGammaShape,tFuture*routingGammaShape/routingGammaMean)    ! cumulative probability at the end of the step
     fractionFuture(iFuture) = max(0._rkind, cumProb - pSave) ! fraction of runoff in the current step
     pSave   = cumProb                                     ! save the cumulative probability for use in the next step
     !write(*,'(a,1x,i4,1x,3(f20.10,1x))') trim(message), iFuture, tFuture, cumProb, fractionFuture(iFuture)
--- a/build/source/engine/var_derive.f90
+++ b/build/source/engine/var_derive.f90
@@ -393,7 +393,6 @@ contains
  real(rkind)                        :: dt                     ! data time step (s)
  integer(i4b)                    :: nTDH                   ! number of points in the time-delay histogram
  integer(i4b)                    :: iFuture                ! index in time delay histogram
- real(rkind)                        :: routingGammaMean       ! mean of the Gamma distribution
  real(rkind)                        :: tFuture                ! future time (end of step)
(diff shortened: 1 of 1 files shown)


andywood (commented):
if routingGammaMean = routingGammaShape * routingGammaScale (on 436) then would
tFuture*routingGammaShape/routingGammaMean  (on 445) = tFuture/routingGammaScale?   (ie simpler)

Also, would routingGammaMean **<=** 0._rkind be appropriate since routingGammaScale could be = 0?

h294liu:
Hi Andy, using "tFuture/routingGammaScale" is indeed simplier and correct as well. routingGammaMean is introduced because it appears in Equation 13 of Clark et al. (2007, WRR). In this way, the code stays consistent with the paper. 
Regarding routingGammaScale, I think mathematically it needs to be positive, cannot be zero. Does this make sense? In summa localParam.txt, its lower and upper limits are 1 and 5000000., respectively.

andywood:
Hi Hongli, sounds good.

I agree that routingGammaScale needs to be >0.  The 'bad values' check
would best trap a user trying to enter 0 (perhaps by mistake, perhaps
because their optimization wasn't good, etc) to avoid the code crashing on
a divide-by-zero error a few lines later.

Btw, gamma is a common statistical distribution, even a fairly
common representation of a unit hydrograph - the implementation could
probably be referenced to any textbook (or not at all) so we're not tied to
a specific paper.   But it's fine to leave it as you did.

Cheers,
-Andy

build/source/engine/var_derive.f90
```diff
@@ -433,16 +433,16 @@ subroutine fracFuture(bpar_data,bvar_data,err,message)
   case(timeDelay)
    ! initialize
    pSave   = 0._rkind ! cumulative probability at the start of the step
-   aLambda = routingGammaShape / routingGammaScale
-   if(routingGammaShape <= 0._rkind .or. aLambda < 0._rkind)then
+   routingGammaMean = routingGammaShape * routingGammaScale
```
martynpclark, line 436: I agree that this `routingGammaMean` can be removed

build/source/engine/var_derive.f90
```diff
@@ -393,7 +393,7 @@ subroutine fracFuture(bpar_data,bvar_data,err,message)
  real(rkind)                        :: dt                     ! data time step (s)
  integer(i4b)                    :: nTDH                   ! number of points in the time-delay histogram
  integer(i4b)                    :: iFuture                ! index in time delay histogram
- real(rkind)                        :: aLambda                ! scale parameter in the Gamma distribution
+ real(rkind)                        :: routingGammaMean       ! mean of the Gamma distribution
```
martynpclark, line 396: I agree that `routingGammaMean` can be removed

build/source/engine/var_derive.f90
```diff
@@ -433,16 +433,16 @@ subroutine fracFuture(bpar_data,bvar_data,err,message)
...
+   if(routingGammaShape <= 0._rkind .or. routingGammaMean < 0._rkind)then
     message=trim(message)//'bad arguments for the Gamma distribution'
     err=20; return
    end if
    ! loop through time steps and compute fraction of runoff in future steps
    do iFuture = 1,nTDH
     ! get weight for a given bin
     tFuture = real(iFuture, kind(dt))*dt                  ! future time (end of step)
-    cumProb = gammp(routingGammaShape,aLambda*tFuture)    ! cumulative probability at the end of the step
+    cumProb = gammp(routingGammaShape,tFuture*routingGammaShape/routingGammaMean)    ! cumulative probability at the end of the step
```
martynpclark, line 445: This line of code can be
```
cumProb = gammp(routingGammaShape,tFuture/routingGammaScale)    ! cumulative probability at the end of the step
```

martynpclark (changes requested):

build/source/engine/var_derive.f90
h294liu: Hi Martyn, the requested changes have been addressed in my second commit. Please take a look. Thanks.

h294liu:
Hi repo managers, I'd like to add a bit more explanations and demos for the changes I made. Hope this eases understanding. 
 
The previous code ACTUALLY has an error in calculating the x value of the Gamma distribution. The reason behind is that, when x value is employed in cumprob = gammp(shape, x), the gammp function requires the x being standardized. Given the two-parameter Gamma distribution, the time delay "tFuture" should be standardized by "tFuture/scale". The old summa code has an error in doing the standardization. Here this issue is fixed by removing variable "aLambda" and updating x value for gammp.

In addition, I add two the figures to show the impacts of code changes on routed runoff. This 2-year simulation is based on the Bow at Banff case study. The first figure shows the summa averageRoutedRunoff of the 1st GRU. The second figure shows the mizuRoute IRFroutedRunoff at the basin outlet. You can see that the new code brings down the peak runoff more or less. 
SUMMA_output
mizuRoute_output

andywood (commented):
This looks fine, we should merge.  Note, it is not 'backward compatible' in the sense that a calibrated SUMMA model in which basin runoff was used for calibration will have changed results.  Users will need to evaluate their upgrade decision for specific cases.  It is backward compatible in the sense that existing model setups will still run without modification.

PR #475, opened 2021-06-30
Sundials as time integrator

zolfaghr:

- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry


build/source/driver/summa_driver.f90
```diff
@@ -89,6 +89,7 @@ program summa_driver
  ! read model forcing data
  call summa_readForcing(modelTimeStep, summa1_struc(n), err, message)
  call handle_err(err, message)
+
  print *, 'step ---> ', iStep
```
wknoben, line 93: This line is not present in the current `develop` branch and prints several outputs to the terminal when the code is run. Is this required or a left-over from debugging?

Example terminal output:
```
1979  6 30  0  0
 step --->         4322
 step --->         4323
 step --->         4324
 step --->         4325
```

martynpclark:
part of the SUNDIALS merge

PR #481, opened 2021-07-24
Fix/restart canopy ice

### LINKED ISSUE #474: Canopy ice content check during initial conditions interferes with restart files under certain conditions
#### Bug Report
- SUMMA v3.0 and above, but presumably has existed for longer
- gcc v7.3.0
- Linux/macOS
- Relevant settings: n/a

Error ```FATAL ERROR: summa_readRestart/check_icond/canopy ice > 0 when canopy temperature > Tfreeze``` can be triggered in cases where a restart file has a non-zero canopy snow/ice state but temperature > Tfreeze. This is not always a physical impossibility in cases when:
- T > Tfreeze when restart file is created and on the next timestep, but canopy ice content has not fully melted (yet)
- T < Tfreeze and canopy ice > 0 when restart file is created and T > Tfreeze on the next timestep

During regular simulations both cases should run fine, but such cases do not make it through a restart. It is not entirely clear to me why this check appears in the ICs and I therefore don't know the best approach to solving this. Thoughts?

wknoben:

- [x] closes #474 (identify the issue associated with this PR)
- [ ] tests passed
- [ ] new tests added
- [x] science test figures
- [x] checked that the new code conforms to the [SUMMA coding conventions](https://github.com/NCAR/summa/blob/master/docs/howto/summa_coding_conventions.md)
- [x] ReleaseNotes entry

Test checkboxes n/a here since this affects restarts only. 

## Issue
Canopy ice content can occasionally go above zero for above-zero temperatures as a result of frost (negative sublimation) appearing on the canopy. In a regular run, this ice typically becomes liquid content on the next timestep (assuming temperature remains above zero). See figure for results from 28 years of half-hourly point-scale simulations, showing that this happens during regular simulations. A restart file created when both `scalarCanopyIce` > 0 and `scalarCanopyTemp` > 0 cannot be used to restart the run, because this condition is found in `check_icond.f90` and the run gracefully exists.

regular_simulation

## Solution
The check is mainly in place to not allow the user to use unrealistic system states as initial conditions. Replacing the current graceful exit with a warning printed to screen retains similar functionality (user is aware of something being off - provided they check the log), while also allowing SUMMA to use its own restart files. In the current implementation, ice content up to 1E-3 is allowed (but flagged in the output log), whereas ice content above this threshold is still a reason for a graceful exit. 

## Test cases
Selected 1 day from the 28-year long run to investigate this issue further. Restart file generated on 2003-01-01 09:00 which matches the critical conditions of `scalarCanopyIce` > 0 and `scalarCanopyTemp` > 0. The short run up to restart file creation is identical to the 28-year run up to that point; see below. At this point:
- `scalarCanopyIce` = 4.76357969e-06 [kg m-2]
- `scalarCanopyTemp` = 276.47829081 [K]

On the following timestep, all remaining ice content becomes liquid and this remains in place until the liquid content goes up further on t = 20:30 h.

create_restart

### Case 1: restart file + regular SUMMA
As expected, the current implementation gracefully doesn't run.

### Case 2: zero canopy ice + regular SUMMA
Setting `scalarCanopyIce` to 0 allows the run to start but leads to differences between the regular (long) run and the restarted run, because this artificially removes some water from the system. See below.

regular_summa_no_canopy_ice

### Case 3: restart file + modified SUMMA
After changes, SUMMA will notify the user that something needs checking:
```
Warning: canopy ice content in restart file (0.4763579693535316E-05) > 0 when canopy temperature > Tfreeze. Continuing.
``` 

but picks up from the restart file and the simulation proceeds as before. See below.

modified_summa_unchanged_restart

The unexpected difference between the long run and the restarted run at t = 20:00 h and t = 20:30 h is possibly related to variable precision. `scalarCanopyLiq` between long run and restarted run differ in order of magnitude of `e-14` to `e-13` before this point (`scalarCanopyIce` values are identical), which may be the cause of these differences. I don't think this is related to the code change but is a more generally related to how restart files are saved and used. 

`scalarCanopyLiq` differences for the period `2003-01-01 09:30` to `2003-01-01 22:00`:
[ 4.83688237e-14],
       [ 7.48117897e-14],
       [ 9.54668710e-14],
       [ 1.09663395e-13],
       [ 1.18571253e-13],
       [ 1.24623781e-13],
       [ 1.28502033e-13],
       [ 1.30901178e-13],
       [ 1.32272187e-13],
       [ 1.33084567e-13],
       [ 1.33609771e-13],
       [ 1.33993237e-13],
       [ 1.34173232e-13],
       [ 1.34276440e-13],
       [ 1.34339499e-13],
       [ 1.34357162e-13],
       [ 1.34346957e-13],
       [ 1.34308945e-13],
       [ 1.34310946e-13],
       [ 1.34325587e-13],
       [ 1.34338748e-13],
       [ 1.87629619e-10],
       [-2.87366370e-09],
       [-7.17519483e-13],
       [-2.52301028e-14],
       [ 2.44843344e-14]


wknoben:
Updated original comment with:
1. Better understanding of what causes the canopy ice content to be > 0 for T > Tfreeze
2. New implementation of the solution that allows small (< 1E-3) ice content values but still exists on ice content larger than this threshold

andywood (changes requested):
I think this looks good, but while you're expanding the informational warning, perhaps go further and write out what the canopy temp, T freeze values are, and identify the canIceTol value as 'canIceTol'.

andywood (commented):
This looks fine after all the back and forth - ok to merge

PR #483, opened 2021-08-12
Add case_study folder & added recent reference to main readme

wknoben:

- [x] ReleaseNotes entry

This adds a `case_study` folder. Purpose is twofold:
- To provide a location to track versioning of certain standard input files (lookup tables, localParamInfo.txt, basinParamInfo.txt)
- To provide users with a hands-on example of a SUMMA setup as part of the repo
 
Test case:
- Albedo decay experiment at Reynolds Mountain East (Figure 6a in the second SUMMA paper; point-scale experiment). 

To be added in due course:
- Mini-catchment example of 2/3 GRUs subdivided into 2/3 HRUs each.


andywood:
Looks fine, though in the 'what's new' doc the reference to 'rk' should be updated to 'rkind'

PR #498, opened 2021-11-10
Doc update: added Dave's installation instructions to MacOS instal doc

bartvanosnabrugge:

- [x ] tests passed (with MkDocs)
- [ x] ReleaseNotes entry

### DIFF
--- a/docs/installation/SUMMA_on_OS_X.md
+++ b/docs/installation/SUMMA_on_OS_X.md
@@ -1,27 +1,32 @@
 # Instructions for installing SUMMA on Mac OS X
 
-> The following has been tested on a MacBook Pro and iMac, both using OS X v. 10.12.6 (Sierra) and with [MacPorts](http://www.macports.org) as the OS X package manager. It should also work on older versions of OS X or with other package managers such as [fink](http://www.finkproject.org) or [homebrew](http://brew.sh), but you will undoubtedly need to make some modifications.
+Two slightly different options have been tested to work. The first uses MacPorts as package manager. The second uses Homebrew as package manager.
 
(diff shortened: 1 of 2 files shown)


docs/installation/SUMMA_on_OS_X.md
```diff
@@ -54,12 +59,68 @@ In the following I will assume that you don't have a Fortran compiler or NetCDF
...
+ INCLUDES=-I/usr/local/include -I/usr/local/opt/lapack/include -I/usr/local/opt/openblas/include
+ LIBRARIES=-L/usr/local/lib -lnetcdff -L/usr/local/opt/lapack/lib -lblas -L/usr/local/opt/openblas/lib -lopenblas
+
+ While Homebrew uses the /usr/local/ dir and this should be consistent, it is recommend to check these folders to verify the files exist.
+
+ Navigating to the SUMMA base directory (i.e. F_MASTER, the root directory of the build folder), run the command
+
+ 6. `make -f build/Makefile.local`
+
+ If the code compiles successfully, then the last line of output from the make process will tell you where the SUMMA executable is installed (it goes into summa/bin). Run summa.exe in that directory (you may need to provide the full path).
```
wknoben, line 126: I asked @arbennett for a second opinion on this and he came up with what's possibly the best solution; merging the general and macOS-specific instructions onto a single page. The macOS instructions would simply be a sub-section on that page. That way there's less hunting around for info and given the similarity between compiling on linux and macOS, users of either system can benefit from seeing instructions for the other (and compiling on Windows isn't possible at all as far as I know, because there's no windows equivalent of the fortran-netcdf library). 

If you're up for it, can you merge this macOS page into the general instructions and tidy the whole thing up a bit?

If that's too much, than I think the general instructions would at least benefit from having a note added that points users to these instructions on the macOS page.

wknoben (changes requested):

martynpclark:
Did not address comments provided.

PR #502, opened 2022-02-17
Fix scalarRainPlusMelt output

h294liu:
The old code outputs scalarRainPlusMelt for the snow+soil layers only, which is incorrect for the no-snow period. The modified code outputs scalarRainPlusMelt for all the soil layers, enabling it works correctly for both the snow and no-snow periods.

Below is a comparison of summa outputs with the old and modified codes for one HRU in August 2000. In the 1st row, RainPlusMelt is almost zero in August which is incorrect because RainPlusMelt is supposed to be equal to Infiltration + SurfaceRunoff. In the 2nd row, the code is modified, and RainPlusMelt =  Infiltration + SurfaceRunoff.

image

Screen Shot 2022-02-17 at 2 38 15 PM

- [ ] tests passed
- [ ] new tests added
- [x] science test figures
- [x] checked that the new code conforms to the [SUMMA coding conventions](https://github.com/NCAR/summa/blob/master/docs/howto/summa_coding_conventions.md)
- [ ] ReleaseNotes entry

### DIFF
--- a/build/source/dshare/flxMapping.f90
+++ b/build/source/dshare/flxMapping.f90
@@ -143,7 +143,7 @@ contains
  flux2state_orig(iLookFLUX%mLayerLiqFluxSnow)               = flux2state(state1=iname_watLayer,  state2=integerMissing)
 
  ! liquid water fluxes for the soil domain
- flux2state_orig(iLookFLUX%scalarRainPlusMelt)              = flux2state(state1=iname_watLayer,  state2=integerMissing)
+ flux2state_orig(iLookFLUX%scalarRainPlusMelt)              = flux2state(state1=iname_matLayer,  state2=integerMissing)
  flux2state_orig(iLookFLUX%scalarMaxInfilRate)              = flux2state(state1=iname_matLayer,  state2=integerMissing)
  flux2state_orig(iLookFLUX%scalarInfiltration)              = flux2state(state1=iname_matLayer,  state2=integerMissing)
  flux2state_orig(iLookFLUX%scalarExfiltration)              = flux2state(state1=iname_matLayer,  state2=integerMissing)
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -19,3 +19,4 @@ This page provides simple, high-level documentation about what has changed in ea
 - Add deflate (compression level) option to outputControl file -- default level is 4 if not specified
 - Fixes an unnecessary rounding error on SAI and LAI values in PHENOLOGY routine
 - Fixes a bug where the SUMMA version is incorrectly reported by "summa.exe -v"
+- Fixes a bug that incorrectly writes scalarRainPlusMelt to output in cases where snow layers do not exist


wknoben:
Hi Hongli, the fix itself looks fine. Could you please:
1. Make sure the PR only contains commits relevant to the issue being fixed;
2. Update the ReleaseNotes entry? 

ReleaseNotes can be found as part of the repo, under `./docs/whats-new.md`

h294liu:
Hi Wouter, I have fixed the two issues you mentioned. Please take a look. Thank you!

andywood:
One question about this -- why does rain plus melt have anything to do with
soil layers (ie "... scalarRainPlusMelt for all the soil layers")?  I'd
have thought it relates only to the snow model and the precipitation inputs
(ie does not include melt from frozen sub-surface moisture).  Perhaps this
was just imprecise language.  In traditional NWS watershed modeling, for
instance, the common variable RAIM (rain + melt) is the melt output from
the snow model plus precipitation that doesn't fall as snow.

The reason I mention it is that RAIM may be of interest if the SUMMA snow
model is run without the soil component, as may happen once it is
BMI-ified.  Perhaps now it is calculated as a sum of infiltration + surface
runoff but it might better be calculated w/o reference to surface fields at
all.

Cheers,
-Andy

h294liu:
Hi Andy, this is an imprecise description. In the old code, scalarRainPlusMelt is output when "state1=iname_watLayer", which means scalarRainPlusMelt is written out when the layers have both snow and soil. This causes scalarRainPlusMelt to be erroneously suppressed for the warm period (eg, Summer) when there is no snow layers, so scalarRainPlusMelt stays zero for the warm period, thought the calculation of scalarRainPlusMelt is correct within summa. 

With "state1=iname_matLayer", the scalarRainPlusMelt is written out as long as the layers have soil.

h294liu:
Considering the condition where there is no soil layer, we might need to change the flux mapping states to be:
`flux2state_orig(iLookFLUX%scalarRainPlusMelt) = flux2state(state1=iname_matLayer,  state2=iname_watLayer)
`
Does this make sense? @wknoben  @martynpclark

wknoben:
Hi Andy,

I think we should put this down to imprecise language yeah. `scalarRainPlusMelt` is the aggregated flux of liquid water (rain + melt) into the soil domain. It doesn't have anything to do with the soil layers as such.

What's meant with this: 
```
The old code outputs scalarRainPlusMelt for the snow+soil layers only, which is incorrect for the no-snow period. 
The modified code outputs scalarRainPlusMelt for all the soil layers, enabling it works correctly for both the snow and no-snow periods.
```
is that previously, the `scalarRainPlusMelt` flux was associated with computations of the snow layer domain only, meaning that its values would only be written to the output file if there were any active snow layers during the timestep (`nSnow` >0). The flux would still be calculated correctly on time steps without a snow layer, but it wouldn't be written to the output file. By associating the flux with the soil domain, under the implicit assumption that there is always a soil domain, the flux will always be written to the output file. This is a similar thing that we encountered with `scalarTotalET` and `scalarNetRadiation` (#487).

I think a further thing that needs to be cleared up is how `scalarRainPlusMelt` is calculated. The `RainPlusMelt = Infiltration + SurfaceRunoff` equation Hongli mentioned is a test that must be true, but it's not have this flux is calculated in the code:  https://github.com/CH-Earth/summa/blob/372c3fbeb3825e3b3d635461a8e552f9f0895aec/build/source/engine/computFlux.f90#L572-L620

As is, no surface fields are used to compute `scalarRainPlusMelt`. Hongli's equation is a re-organization of what happens in `soilLiqFlx.f90`: https://github.com/CH-Earth/summa/blob/372c3fbeb3825e3b3d635461a8e552f9f0895aec/build/source/engine/soilLiqFlx.f90#L1345-L1349

I'm admittedly not very familiar with the output-writing logic but it seems to me there are two possible cases:
1. In cases where no soil computations are run, but a soil domain has been defined, I think `scalarRainPlusMelt` would end up correctly in the output file;
2. In cases where no soil domain exists at all (is this possible in the current code?), I have no idea what would happen w.r.t. output writing of variables associated with the soil domain.

The proposed solution to associate the flux with both the snow layers and the soil layers separately:
```
 flux2state_orig(iLookFLUX%scalarRainPlusMelt) = flux2state(state1=iname_matLayer, state2=iname_watLayer) 
```
could still break in cases where no soil domain has been defined and no snow layers exist on a given time step, depending on what happens in case (2) above. I'd suggest to keep this in the back of our minds until we actually are at a stage where we're trying to run these no-soil cases (or run those tests now and see what happens, before shooting from the hip). Thoughts?

andywood:
Thanks for breaking it down like this.  There should always be a rain+melt
value regardless of presence of snow (and without reference to surface
fluxes like infiltration or soil states, or association with the soil
domain code except as intent(in)).  I think we just have to make sure it's
assigned and updated in each of the model spaces excluding subsurface.
That said we've made mistakes before on mass balance variables, so double
checking is a good idea.

andywood:
Reminded by the new PR comment -- did we ever check that scalarRainPlusMelt actually does equal rain + melt?  ie liquid precipitation + snowmelt?

PR #508, opened 2022-04-01
update scalarSnowDrainage

guoqiang-tang:

1. `scalarSnowDrainage=drainageMeltPond/iden_water` is added when` nSnow==0` according to Martyn's suggestion.
2. The localParamInfo.txt in the case_study is updated. Compared with the old localParamInfo.txt, the new localParamInfo.txt has several parameter initial and bound values changes made by Andy Wood, which can help achieve better model performances. Besides, I changed the upper bounds of k_soil and heightCanopyBottom explained as below. The increased upper bounds allow appropriate parameter range in calibration.
(1) In TBL_SOILPARM.TBL (Soil Parameters: ROSETTA), the k_soil is up to 7.43852e-05 for 'SAND'. Therefore, **the upper bound of k_soil is increased from 100.d-07 to 1.d-04** in the localParamInfo.txt.
(2) In TBL_VEGPARM.TBL, for MODIFIED_IGBP_MODIS_NOAH, ZBOTV is up to 11.5 for 'Deciduous Broadleaf Forest'; for USGS, ZBOTV is up to 11.5 for 'Deciduous Broadleaf Forest' and 'Wooded Wetland'. Therefore, **the upper bound of heightCanopyBottom is increased from 5.0000 to 15.000** in the localParamInfo.txt.

### DIFF
--- a/build/source/engine/computFlux.f90
+++ b/build/source/engine/computFlux.f90
@@ -613,8 +613,9 @@ contains
   ! define forcing for the soil domain for the case of no snow layers
   ! NOTE: in case where nSnowOnlyHyd==0 AND snow layers exist, then scalarRainPlusMelt is taken from the previous flux evaluation
   if(nSnow==0)then
+   scalarSnowDrainage=drainageMeltPond/iden_water ! melt of the snow without a layer (m s-1)
    scalarRainPlusMelt = (scalarThroughfallRain + scalarCanopyLiqDrainage)/iden_water &  ! liquid flux from the canopy (m s-1)
-                         + drainageMeltPond/iden_water  ! melt of the snow without a layer (m s-1)
+                         + scalarSnowDrainage
   endif  ! if no snow layers
 
  endif
--- a/case_study/base_settings/localParamInfo.txt
+++ b/case_study/base_settings/localParamInfo.txt
@@ -1,3 +1,8 @@
+! File provenance
+! ---------------
+! 2021-08-10: taken from the SUMMA test cases v3 distribution (WRR figures) and included as case study default file
+!
+!
 ! =======================================================================================================================
 ! =======================================================================================================================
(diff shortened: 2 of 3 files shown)


wknoben:
Thanks for the PR. Could you please address the following?
- Address the comment about the parameter file header.
- Separate the changes to the code from changes to the parameter files into their own PRs. This keeps the history clean and means that, if we have to revert one of these changes, the other one isn't affected.

@andywood could you please have a look at the changes to the numerical settings in the parameter file? If you can recall why these changes were needed it'd be very helpful if you could outline that here so we can refer to the info later, in case that becomes necessary.

andywood:
The localParam changes look fine to me ... I'm not sure where the various versions of these came from, tbh.  My current localParam settings are not the same as either the old or new versions here, though they match in some places.  At some point I recall the range on k_macropore to reach higher values, I think.  I support the min canopy height max bound going higher, maybe even more than 15m -- e.g., think Douglas fir ...
https://www.nps.gov/articles/images/MORA-DouglasFirForest_Longmire-20171013.jpg?maxwidth=1200&autorotate=false

There may end up being different versions of localParams (& other params/attributes) associated with different test cases, but it would be good to document changes to the main defaults included with the code.  This would particularly highlight where the change was a 'fix' for some poor behavior.  

I'm not sure code & param file changes should always be separated, especially in the case where code may be added that requires a new parameter, in which case it would be odd to put that into a different PR.  But I agree on the documentation aspect, and if the localParam update is totally unconnected from the rest of the PR, no reason to combine them.

tgq14:
I agree with the discussion about localParam. At least for this PR, the code change and localParam change should be separated. I will limit this PR to scalarSnowDrainage and create a localParam PR later.

martynpclark:
Thanks Guoqiang, included in a separate PR

PR #511, opened 2022-04-19
specify velocity parameter to calculate scale of the time delay histo…

martynpclark:
Initial capabilities to use a velocity parameter to define the scale of the Gamma distribution. Hongli will complete testing.

### DIFF
--- a/build/source/driver/summa_setup.f90
+++ b/build/source/driver/summa_setup.f90
@@ -255,12 +255,6 @@ contains
  ! loop through GRUs
  do iGRU=1,nGRU
 
-  ! calculate the fraction of runoff in future time steps
-  call fracFuture(bparStruct%gru(iGRU)%var,    &  ! vector of basin-average model parameters
-                  bvarStruct%gru(iGRU),        &  ! data structure of basin-average variables
-                  err,cmessage)                   ! error control
-  if(err/=0)then; message=trim(message)//trim(cmessage); return; endif
-
   ! loop through local HRUs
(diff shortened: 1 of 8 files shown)


martynpclark:
@h294liu -- returning to an old pull request. Can you provide an update on the progress of these developments?

h294liu:
I am so sorry for not following up on this work! Sure, I am happy to
complete the test after submitting the pyviscous manuscript this week. Does
this work?

martynpclark:
Yes, all is perfect! Thanks so much Hongli!!!

h294liu (commented):
I have tested and found that SUMMA run through these new codes without any problems. Yeah! 

---
**Test Setting Configurations**

- Set subRouting as _constVel_ in the "modelDecisions.txt" file.
- Add the following to "basinParamInfo.txt" file: _routingVelocity           |       1.00 |       0.1000 |     100.0000_
Note that the three numbers are the default, min, and max values of routingVelocity.


--- 
**Test Results**
The test domain is the _Bow at Banff_ basin. The model includes 51 GRUs and runs for the year of 2008 (365 days). The model output is daily _averageRoutedRunoff_mean_. 

Below is the model output comparison for one GRU (10 days). 

- The black line is for the _timeDlay_ method (old method). 
- The other two lines are for the _constVel_ method (newly added method). 
- The red line uses flow velocity=100m/s, and the blue line uses  flow velocity=0.1m/s.

Screenshot 2023-08-03 at 10 31 51 PM

PR #525, opened 2023-05-24
Refactoring flux files

seantrim:

Refactoring of the flux related files vegNrgFlux.f90, ssdNrgFlux.f90, vegLiqFlux.f90, snowLiqFlux.f90, soilLiqFlux.f90, groundwatr.f90, bigAquifer.f90, and computFlux.f90 was performed. This includes cleanup, shortening, and  updates to conform to the modern Fortran standard. Overall, the flux files have been shortened by nearly 900 lines. Updates to Fortran arithmetic expressions were also made to improve accuracy while reducing computational cost (such as using integer exponents and the intrinsic sqrt function where possible). There was also a bug with the initialization of C_r in vegNrgFlux.f90 that has been fixed (C_r was unintentionally assigned a literal of default real type -- it is now assigned a value of type rkind). Additionally, a script has been added to simplify the compilation process of SUMMA (summa/build/compile_SUMMA.sh), which was helpful when recompiling during the testing process.

The laugh tests were used to confirm functionality of the proposed changes. Apart from the Wigmosta tests, all laugh tests give bit identical results compared to the original develop branch. For the Wigmosta tests, the change in variable output does not exceed $10^{-7}$. This change is due to reduced round off error in several arithmetic expressions.

### DIFF
--- a/build/source/engine/bigAquifer.f90
+++ b/build/source/engine/bigAquifer.f90
@@ -108,7 +108,7 @@ contains
  scalarAquiferBaseflow = aquiferBaseflowRate*(xTemp**aquiferBaseflowExp)
 
  ! compute the derivative in the net aquifer flux
- dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1_i4b)))/aquiferScaleFactor ! aquiferBaseflowExp is integer valued
+ dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1._rkind)))/aquiferScaleFactor
 
  end associate ! end association to data in structures
 
--- a/build/source/engine/bigAquifer.f90
+++ b/build/source/engine/bigAquifer.f90
@@ -20,7 +20,6 @@
 
 module bigAquifer_module
 ! -----------------------------------------------------------------------------------------------------------
-
 ! numerical recipes data types
 USE nrtype
 
@@ -30,15 +29,13 @@ USE globalData,only:realMissing     ! missing real number
 
 ! physical constants
 USE multiconst,only:&
-                    LH_vap,  & ! latent heat of vaporization   (J kg-1)
-                    iden_water ! intrinsic density of water    (kg m-3)
-
+                    LH_vap,  &      ! latent heat of vaporization   (J kg-1)
+                    iden_water      ! intrinsic density of water    (kg m-3)
 ! -----------------------------------------------------------------------------------------------------------
 implicit none
 private
-public::bigAquifer
+public :: bigAquifer
 contains
-
 
  ! ***************************************************************************************************************
  ! public subroutine soilLiqFlx: compute liquid water fluxes and their derivatives
@@ -59,10 +56,10 @@ contains
                        ! output: error control
                        err,message)                    ! intent(out): error control
  ! named variables
- USE var_lookup,only:iLookDIAG              ! named variables for structure elements
- USE var_lookup,only:iLookPARAM             ! named variables for structure elements
+ USE var_lookup,only:iLookDIAG                         ! named variables for structure elements
+ USE var_lookup,only:iLookPARAM                        ! named variables for structure elements
  ! data types
- USE data_types,only:var_dlength            ! x%var(:)%dat   (dp)
+ USE data_types,only:var_dlength                       ! x%var(:)%dat   (dp)
  ! -------------------------------------------------------------------------------------------------------------------------------------------------
  implicit none
  ! input: state variables, fluxes, and parameters
@@ -70,23 +67,22 @@ contains
  real(rkind),intent(in)              :: scalarCanopyTranspiration    ! canopy transpiration (kg m-2 s-1)
  real(rkind),intent(in)              :: scalarSoilDrainage           ! soil drainage (m s-1)
  ! input: diagnostic variables and parameters
- type(var_dlength),intent(in)     :: mpar_data                    ! model parameters
- type(var_dlength),intent(in)     :: diag_data                    ! diagnostic variables for a local HRU
+ type(var_dlength),intent(in)        :: mpar_data                    ! model parameters
+ type(var_dlength),intent(in)        :: diag_data                    ! diagnostic variables for a local HRU
  ! output: fluxes
  real(rkind),intent(out)             :: scalarAquiferTranspire       ! transpiration loss from the aquifer (m s-1)
  real(rkind),intent(out)             :: scalarAquiferRecharge        ! recharge to the aquifer (m s-1)
  real(rkind),intent(out)             :: scalarAquiferBaseflow        ! total baseflow from the aquifer (m s-1)
  real(rkind),intent(out)             :: dBaseflow_dAquifer           ! change in baseflow flux w.r.t. aquifer storage (s-1)
  ! output: error control
- integer(i4b),intent(out)         :: err                          ! error code
- character(*),intent(out)         :: message                      ! error message
+ integer(i4b),intent(out)            :: err                          ! error code
+ character(*),intent(out)            :: message                      ! error message
  ! -----------------------------------------------------------------------------------------------------------------------------------------------------
  ! local variables
  real(rkind)                         :: aquiferTranspireFrac         ! fraction of total transpiration that comes from the aquifer (-)
  real(rkind)                         :: xTemp                        ! temporary variable (-)
  ! -------------------------------------------------------------------------------------------------------------------------------------------------
- ! initialize error control
- err=0; message='bigAquifer/'
+ err=0; message='bigAquifer/' ! initialize error control
 
  ! make association between local variables and the information in the data structures
  associate(&
@@ -98,7 +94,7 @@ contains
  aquiferBaseflowRate    => mpar_data%var(iLookPARAM%aquiferBaseflowRate)%dat(1),   & ! intent(in): [dp] tbaseflow rate when aquiferStorage = aquiferScaleFactor (m s-1)
  aquiferScaleFactor     => mpar_data%var(iLookPARAM%aquiferScaleFactor)%dat(1),    & ! intent(in): [dp] scaling factor for aquifer storage in the big bucket (m)
  aquiferBaseflowExp     => mpar_data%var(iLookPARAM%aquiferBaseflowExp)%dat(1)     & ! intent(in): [dp] baseflow exponent (-)
- )  ! associating local variables with the information in the data structures
+ )  ! end associating local variables with the information in the data structures
 
  ! compute aquifer transpiration (m s-1)
  aquiferTranspireFrac   = scalarAquiferRootFrac*scalarTranspireLimAqfr/scalarTranspireLim   ! fraction of total transpiration that comes from the aquifer (-)
@@ -112,16 +108,10 @@ contains
  scalarAquiferBaseflow = aquiferBaseflowRate*(xTemp**aquiferBaseflowExp)
 
  ! compute the derivative in the net aquifer flux
- dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1._rkind)))/aquiferScaleFactor
+ dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1_i4b)))/aquiferScaleFactor ! aquiferBaseflowExp is integer valued
 
- ! end association to data in structures
- end associate
+ end associate ! end association to data in structures
 
  end subroutine bigAquifer
-
-
- ! *******************************************************************************************************************************************************************************
- ! *******************************************************************************************************************************************************************************
-
 
 end module bigAquifer_module
--- a/build/source/engine/groundwatr.f90
+++ b/build/source/engine/groundwatr.f90
@@ -24,7 +24,7 @@ module groundwatr_module
 USE nrtype
 
 ! model constants
-USE multiconst,only:iden_water ! density of water (kg m-3)
+USE multiconst,only:iden_water   ! density of water (kg m-3)
 
 ! derived types to define the data structures
 USE data_types,only:&
@@ -40,9 +40,9 @@ USE var_lookup,only:iLookPARAM   ! named variables for structure elements
 
 ! look-up values for the choice of groundwater parameterization
 USE mDecisions_module,only:  &
- qbaseTopmodel,              & ! TOPMODEL-ish baseflow parameterization
- bigBucket,                  & ! a big bucket (lumped aquifer model)
- noExplicit                    ! no explicit groundwater parameterization
+ qbaseTopmodel,              &   ! TOPMODEL-ish baseflow parameterization
+ bigBucket,                  &   ! a big bucket (lumped aquifer model)
+ noExplicit                      ! no explicit groundwater parameterization
 
 ! privacy
 implicit none
@@ -51,7 +51,7 @@ real(rkind),parameter     :: valueMissing=-9999._rkind    ! missing value parame
 real(rkind),parameter     :: verySmall=epsilon(1.0_rkind) ! a very small number (used to avoid divide by zero)
 real(rkind),parameter     :: dx=1.e-8_rkind               ! finite difference increment
 private
-public::groundwatr
+public :: groundwatr
 contains
 
 
@@ -66,7 +66,7 @@ contains
  ! "field capacity" below the depth zCrit, where zCrit is defined as the lowest point in the soil
  ! profile where the volumetric liquid water content is less than field capacity.
  !
- ! We further assume that transmssivity (m2 s-1) for each layer is defined asuming that the water
+ ! We further assume that transmssivity (m2 s-1) for each layer is defined assuming that the water
  ! available for saturated flow is located at the bottom of the soil profile. Specifically:
  !  trTotal(iLayer) = tran0*(zActive(iLayer)/soilDepth)**zScale_TOPMODEL
  !  trSoil(iLayer)  = trTotal(iLayer) - trTotal(iLayer+1)
@@ -79,31 +79,26 @@ contains
  !
  ! ************************************************************************************************
  subroutine groundwatr(&
-
                        ! input: model control
                        nSnow,                                  & ! intent(in): number of snow layers
                        nSoil,                                  & ! intent(in): number of soil layers
                        nLayers,                                & ! intent(in): total number of layers
                        getSatDepth,                            & ! intent(in): logical flag to compute index of the lowest saturated layer
-
                        ! input: state and diagnostic variables
                        mLayerdTheta_dPsi,                      & ! intent(in): derivative in the soil water characteristic w.r.t. matric head in each layer (m-1)
                        mLayerMatricHeadLiq,                    & ! intent(in): liquid water matric potential (m)
                        mLayerVolFracLiq,                       & ! intent(in): volumetric fraction of liquid water (-)
                        mLayerVolFracIce,                       & ! intent(in): volumetric fraction of ice (-)
-
                        ! input/output: data structures
                        attr_data,                              & ! intent(in):    spatial attributes
                        mpar_data,                              & ! intent(in):    model parameters
                        prog_data,                              & ! intent(in):    model prognostic variables for a local HRU
                        diag_data,                              & ! intent(in):    model diagnostic variables for a local HRU
                        flux_data,                              & ! intent(inout): model fluxes for a local HRU
-
                        ! output: baseflow
                        ixSaturation,                           & ! intent(inout) index of lowest saturated layer (NOTE: only computed on the first iteration)
                        mLayerBaseflow,                         & ! intent(out): baseflow from each soil layer (m s-1)
                        dBaseflow_dMatric,                      & ! intent(out): derivative in baseflow w.r.t. matric head (s-1)
-
                        ! output: error control
                        err,message)                              ! intent(out): error control
  ! ---------------------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


wknoben:
Hi Sean,

Thanks for this. Three things:

1. Can you confirm that you ran the full suite of SUMMA test cases (https://www.hydroshare.org/resource/e4df3065264341a48017f6d2f4ad030c/) and not just the laugh tests (https://github.com/CH-Earth/laughTests)? The full suite covers more of the model behavior than the laugh tests do. 

2. Can you add an example (specific lines that were changed and a plot of a relevant flux or state variable) for the Wigmosta case? I don't doubt that you're right but it would be good to have an accurate record of which changes in the code caused these changes in output in case we ever need to get back to it. Also, how do we know round off error is reduced in the new implementation?

3. W.r.t the compile script, does that need any additional comments to indicate (1) to novice users that this is not fully automatic but may need manual updates to library paths, and (2) what the prerequisites for use are?

wknoben:
Thanks for the updates. 

1. I noticed that the discussion on `sqrt()` vs `**0.5` refers to R instead of Fortran, though playing around with a test function in an online Fortran environment seems to confirm that `sqrt()` is considerably faster than `**0.5` (3 tests each, execution time 7.70, 7.86, 7.79 for `0.5`; 2.14, 2.13, 2.21 for `sqrt()`).

<details><summary>Code</summary>

```
Program test_sqrt_methods

implicit none
integer :: N,i
real :: T1,T2,x,y

! initiate time
call cpu_time(T1)

! Run square root
N = 1000000000
x = 10
do i = 1,N
    !y = x**(0.5)
    y = sqrt(x)
end do

! End time
call cpu_time(T2)
print *, T1,T2, T2-T1

End Program test_sqrt_methods
```
Tested here: https://www.onlinegdb.com/online_fortran_compiler

</details>

2. I'll have a look at the code changes and leave specific comments if I see anything. Given what kind of work this is it'd be good if @martynpclark can approve these too.

3. Last, for future work it might make sense to separate cosmetic changes and code changes into separate PRs. That'll make it easier to see those changes that fundamentally affect model simulations.

build/source/engine/bigAquifer.f90
```diff
@@ -108,7 +108,7 @@ subroutine bigAquifer(&
  scalarAquiferBaseflow = aquiferBaseflowRate*(xTemp**aquiferBaseflowExp)
 
  ! compute the derivative in the net aquifer flux
- dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1_i4b)))/aquiferScaleFactor ! aquiferBaseflowExp is integer valued
+ dBaseflow_dAquifer    = -(aquiferBaseflowExp*aquiferBaseflowRate*(xTemp**(aquiferBaseflowExp - 1._rkind)))/aquiferScaleFactor
```
wknoben, line 111: Why does this revert the initial change from `rkind` to `i4b`?

This one is the first such a reverted change that I see. Are there more?

build/source/engine/groundwatr.f90
```diff
@@ -225,70 +209,7 @@ subroutine groundwatr(&
...
-  ! print the numerical Jacobian
-  print*, '** numerical Jacobian:'
-  write(*,'(a4,1x,100(i12,1x))') 'xCol', (iLayer, iLayer=1,nSoil)
-  do iLayer=1,nSoil; write(*,'(i4,1x,100(e12.5,1x))') iLayer, nJac(1:nSoil,iLayer); end do
-  !pause 'testing Jacobian'
-
- end if  ! if desire to compute the Jacobian
-
- ! end association to variables in data structures
- end associate
```
wknoben, line 291: Why can this section be taken out?

build/source/engine/groundwatr.f90
```diff
@@ -419,117 +328,81 @@ subroutine computeBaseflow(&
...
- ! compute the exfiltartion (m s-1)
+ ! compute the exfiltration (m s-1)
  if(totalColumnInflow > totalColumnOutflow .and. logF > tiny(1._rkind))then
   scalarExfiltration = logF*(totalColumnInflow - totalColumnOutflow)  ! m s-1
-  !write(*,'(a,1x,10(f30.20,1x))') 'scalarExfiltration = ', scalarExfiltration
  else
   scalarExfiltration = 0._rkind
  end if
 
- ! check
```
wknoben, line 463: @martynpclark Can you give an opinion on whether it's worth keeping (informal) debugging statements like these?

build/source/engine/groundwatr.f90
```diff
@@ -419,117 +328,81 @@ subroutine computeBaseflow(&
...
 
- ! compute the total baseflow
- qbTotal = sum(mLayerBaseflow)
+ qbTotal = sum(mLayerBaseflow) ! compute the total baseflow
 
  ! add exfiltration to the baseflow flux at the top layer
  mLayerBaseflow(1)      = mLayerBaseflow(1) + scalarExfiltration
  mLayerColumnOutflow(1) = mLayerColumnOutflow(1) + scalarExfiltration*HRUarea
 
- ! test
```
wknoben, line 478: @martynpclark Can you also give an opinion on whether it's worth keeping (formal) debugging statements like these?

build/source/engine/groundwatr.f90
```diff
@@ -419,117 +328,81 @@ subroutine computeBaseflow(&
...
-  xTran  = tran0*(xDepth/soilDepth)**zScale_TOPMODEL  ! transmissivity for the entire aquifer (m2 s-1)
-  xFlow  = xTran*tan_slope*contourLength/HRUarea   ! total column outflow (m s-1)
-  print*, 'ixSaturation = ', ixSaturation
-  write(*,'(a,1x,5(f30.20,1x))') 'tran0, soilDepth                 = ', tran0, soilDepth
-  write(*,'(a,1x,5(f30.20,1x))') 'surfaceHydCond, zScale_TOPMODEL  = ', surfaceHydCond, zScale_TOPMODEL
-  write(*,'(a,1x,5(f30.20,1x))') 'xDepth, zActive(ixSaturation)    = ', xDepth, zActive(ixSaturation)
-  write(*,'(a,1x,5(f30.20,1x))') 'xTran, trTotal(ixSaturation)     = ', xTran, trTotal(ixSaturation)
-  write(*,'(a,1x,5(f30.20,1x))') 'xFlow, totalColumnOutflow        = ', xFlow, sum(mLayerColumnOutflow(:))/HRUarea
-  !pause 'check groundwater'
- end if
```
wknoben, line 490: @martynpclark Can you also give an opinion on whether it's worth keeping (formal) debugging statements like these?

build/source/engine/snowLiqFlx.f90
```diff
@@ -128,7 +128,7 @@ subroutine snowLiqFlx(&
  end if
 
  ! check the meltwater exponent is >=1
- if (mw_exp<1_i4b) then; err=20; message=trim(message)//'meltwater exponent < 1'; return; end if ! mw_exp is an i4b integer
+ if (mw_exp<1._rkind) then; err=20; message=trim(message)//'meltwater exponent < 1'; return; end if
```
wknoben, line 131: Another `i4b` to `rkind` change

build/source/engine/snowLiqFlx.f90
```diff
@@ -177,7 +177,7 @@ subroutine snowLiqFlx(&
    availCap  = mLayerPoreSpace(iLayer) - mLayerThetaResid(iLayer)                 ! available capacity
    relSaturn = (mLayerVolFracLiqTrial(iLayer) - mLayerThetaResid(iLayer)) / availCap    ! relative saturation
    iLayerLiqFluxSnow(iLayer)      = k_snow*relSaturn**mw_exp
-   iLayerLiqFluxSnowDeriv(iLayer) = ( (k_snow*mw_exp)/availCap ) * relSaturn**(mw_exp - 1_i4b) ! mw_exp is an i4b integer
+   iLayerLiqFluxSnowDeriv(iLayer) = ( (k_snow*mw_exp)/availCap ) * relSaturn**(mw_exp - 1._rkind)
```
wknoben, line 180: Another `i4b` to `rkind` change

build/source/engine/soilLiqFlx.f90
```diff
@@ -671,86 +644,61 @@ subroutine soilLiqFlx(&
...
-  !                                                  iLayer, vectorMatricHeadTrial, iLayerHydCond(iLayer), iLayerLiqFluxSoil(iLayer)
-  !if(iLayer==1) write(*,'(a,i4,1x,L1,1x,2(e15.5,1x))') 'iLayer, (ixDerivMethod==numerical), dq_dHydStateBelow(iLayer-1), dq_dHydStateAbove(iLayer) = ', &
-  !                                                      iLayer, (ixDerivMethod==numerical), dq_dHydStateBelow(iLayer-1), dq_dHydStateAbove(iLayer)
-  !pause
-
- end do  ! (looping through soil layers)
-
- ! add infiltration to the upper-most unfrozen layer
- ! NOTE: this is done here rather than in surface runoff
- !iLayerLiqFluxSoil(ixIce) = iLayerLiqFluxSoil(ixIce) + scalarSurfaceInfiltration
```
wknoben, line 693: This seems like legacy code. 
Can we check that this operation (adding infiltration to upper-most unfrozen layer) is in fact done somewhere before we remove this code?

build/source/engine/soilLiqFlx.f90
```diff
@@ -1015,71 +937,47 @@ subroutine diagv_node(&
...
-     effSat = (volLiq - theta_res)/(theta_sat - volIce - theta_res)
-     psiLiq = matricHead(effSat,vGn_alpha,0._rkind,1._rkind,vGn_n,vGn_m)  ! use effective saturation, so theta_res=0 and theta_sat=1
-     hydCon = hydCond_psi(psiLiq,scalarSatHydCond,vGn_alpha,vGn_n,vGn_m)
-     call iceImpede(volIce,f_impede,iceImpedeFac,dIceImpede_dLiq)
-     hydIce = hydCon*iceImpedeFac
-     print*, 'test derivative: ', (psiLiq - scalarMatricHeadTrial)/dx, dPsiLiq_dTemp
-     print*, 'test derivative: ', (hydCon - hydCond_noIce)/dx, dHydCondMicro_dTemp
-     print*, 'test derivative: ', (hydIce - scalarHydCond)/dx, dHydCond_dTemp
-     print*, 'press any key to continue'; read(*,*) ! (alternative to the deprecated 'pause' statement)
-    end if  ! testing the derivative
```
wknoben, line 1063: More old testing code. Worth keeping? @martynpclark

build/source/engine/soilLiqFlx.f90
```diff
@@ -1263,112 +1157,76 @@ subroutine surfaceFlx(&
...
-   !print*, 'scalarRainPlusMelt, xMaxInfilRate = ', scalarRainPlusMelt, xMaxInfilRate
-   !print*, 'scalarSurfaceInfiltration, scalarSurfaceRunoff = ', scalarSurfaceInfiltration, scalarSurfaceRunoff
-   !print*, '(1._rkind - scalarFrozenArea), (1._rkind - scalarFrozenArea)*scalarInfilArea = ', (1._rkind - scalarFrozenArea), (1._rkind - scalarFrozenArea)*scalarInfilArea
+   scalarSurfaceInfiltration = (1._rkind - scalarFrozenArea)*scalarInfilArea*min(scalarRainPlusMelt,xMaxInfilRate) ! compute infiltration (m s-1)
+   scalarSurfaceRunoff = scalarRainPlusMelt - scalarSurfaceInfiltration ! compute surface runoff (m s-1)
 
    ! set surface hydraulic conductivity and diffusivity to missing (not used for flux condition)
-   surfaceHydCond = realMissing
-   surfaceDiffuse = realMissing
+   surfaceHydCond = realMissing; surfaceDiffuse = realMissing
```
wknoben, line 1219: Might be cleaner to keep these on separate lines

build/source/engine/soilLiqFlx.f90
```diff
@@ -1263,112 +1157,76 @@ subroutine surfaceFlx(&
...
    ! NOTE 1: Depends on multiple soil layers and does not jive with the current tridiagonal matrix
    ! NOTE 2: Need to define the derivative at every call, because intent(out)
-   dq_dHydState = 0._rkind
-   dq_dNrgState = 0._rkind
-
-  ! ***** error check
-  case default; err=20; message=trim(message)//'unknown upper boundary condition for soil hydrology'; return
-
- end select  ! (type of upper boundary condition)
+   dq_dHydState = 0._rkind; dq_dNrgState = 0._rkind
```
wknoben, line 1224: Might be cleaner to keep these on separate lines

build/source/engine/vegNrgFlux.f90
```diff
@@ -581,70 +519,58 @@ subroutine vegNrgFlux(&
...
- ! initialize error control
- err=0; message="vegNrgFlux/"
+ err=0; message="vegNrgFlux/" ! initialize error control
 
- ! initialize printflag
- printflag = .false.
-
- ! identify the type of boundary condition for thermodynamics
- select case(ix_bcUpprTdyn)
+ printflag = .false. ! initialize printflag
```
wknoben, line 543: If (big if) we remove all the debug printing statements that use `printFlag`, should we delete it here too or does it serve multiple purposes?

build/source/engine/vegNrgFlux.f90
```diff
@@ -655,119 +581,93 @@ subroutine vegNrgFlux(&
...
-      ! compute diffuse transmissivity (-)
-      diffuseTrans = (1._rkind - scaleLAI)*exp(-scaleLAI) + (scaleLAI**2._rkind)*expi
-      ! compute the canopy emissivity
-      scalarCanopyEmissivity = (1._rkind - diffuseTrans)*vegEmissivity
-     ! *** check we found the correct option
-     case default
+     case(difTrans) ! *** canopy emissivity parameterized as a function of diffuse transmissivity
+      scaleLAI = 0.5_rkind*exposedVAI  ! compute the exponential integral
+      expi     = expInt(scaleLAI)      ! compute the exponential integral
+      diffuseTrans = (1._rkind - scaleLAI)*exp(-scaleLAI) + (scaleLAI**2_i4b)*expi ! compute diffuse transmissivity (-)
```
wknoben, line 651: If we're swapping `rkind` to `i4b` in exponents to avoid roundoff errors, should we do the same for these kind of constants? `(1._rkind - scaleLAI)`

build/source/engine/vegNrgFlux.f90
```diff
@@ -1148,37 +991,26 @@ subroutine vegNrgFlux(&
...
-     ! (resistance from the soil [s m-1])
-     scalarSoilResistance = scalarGroundSnowFraction*1._rkind + (1._rkind - scalarGroundSnowFraction)*EXP(8.25_rkind - 4.225_rkind*soilEvapFactor)  ! Sellers (1992)
-     !scalarSoilResistance = scalarGroundSnowFraction*0._rkind + (1._rkind - scalarGroundSnowFraction)*exp(8.25_rkind - 6.0_rkind*soilEvapFactor)    ! Niu adjustment to decrease resitance for wet soil
-     ! (relative humidity in the soil pores [0-1])
-     if(mLayerMatricHead(1) > -1.e+6_rkind)then  ! avoid problems with numerical precision when soil is very dry
-      soilRelHumidity_noSnow = exp( (mLayerMatricHead(1)*gravity) / (groundTemp*R_wv) )
+    if (firstFluxCall) then
+     soilEvapFactor = mLayerVolFracLiq(nSnow+1)/(theta_sat - theta_res) ! soil water evaporation factor [0-1]
+     ! resistance from the soil [s m-1] -- Sellers (1992)
+     scalarSoilResistance = scalarGroundSnowFraction + (1._rkind - scalarGroundSnowFraction)*exp(8.25_rkind - 4.225_rkind*soilEvapFactor) ! removed factor of unity
```
wknoben, line 1005: The comment `! removed factor of unity` only makes sense in the context of the now removed factor. Might be better to add the original equation in the comment instead:

```
scalarSoilResistance = scalarGroundSnowFraction + (1._rkind - scalarGroundSnowFraction)*exp(8.25_rkind - 4.225_rkind*soilEvapFactor) ! Full equation (removed factor of unity for speed): scalarSoilResistance = scalarGroundSnowFraction*1._rkind + (1._rkind - scalarGroundSnowFraction)*EXP(8.25_rkind - 4.225_rkind*soilEvapFactor)
```

build/source/engine/vegNrgFlux.f90
```diff
@@ -1148,37 +991,26 @@ subroutine vegNrgFlux(&
...
+     if (mLayerMatricHead(1) > -1.e+6_rkind) then  ! avoid problems with numerical precision when soil is very dry
+      soilRelHumidity_noSnow = exp((mLayerMatricHead(1)*gravity)/(groundTemp*R_wv))
      else
       soilRelHumidity_noSnow = 0._rkind
-     end if ! (if matric head is very low)
-     scalarSoilRelHumidity  = scalarGroundSnowFraction*1._rkind + (1._rkind - scalarGroundSnowFraction)*soilRelHumidity_noSnow
-     !print*, 'mLayerMatricHead(1), scalarSoilRelHumidity = ', mLayerMatricHead(1), scalarSoilRelHumidity
-    end if  ! (if the first flux call)
+     end if ! if matric head is very low
+     scalarSoilRelHumidity  = scalarGroundSnowFraction + (1._rkind - scalarGroundSnowFraction)*soilRelHumidity_noSnow ! removed factor of unity
```
wknoben, line 1012: As above

build/source/engine/vegNrgFlux.f90
```diff
@@ -2070,104 +1760,92 @@ subroutine aeroResist(&
...
-    approxDragCoef   = min( sqrt(C_s + C_r*exposedVAI/2._rkind), approxDragCoef_max)
+    ! coupute roughness length of the veg canopy
+    approxDragCoef   = min(sqrt(C_s + C_r*exposedVAI/2._rkind), approxDragCoef_max)
     z0Canopy         = (1._rkind - fracCanopyHeight) * exp(-vkc*approxDragCoef - psi_h) * (heightCanopyTopAboveSnow-heightCanopyBottomAboveSnow)
-
    ! Choudhury and Monteith (QJRMS 1988) "A four layer model for the heat budget..."
    case(CM_QJRMS1988)
     funcLAI =  cd_CM*exposedVAI
-    zeroPlaneDisplacement = 1.1_rkind*heightCanopyTopAboveSnow*log(1._rkind + funcLAI**0.25_rkind)
+    zeroPlaneDisplacement = 1.1_rkind*heightCanopyTopAboveSnow*log(1._rkind + sqrt(sqrt(funcLAI))) ! using sqrt intrinsic for speed
```
wknoben, line 1835: This new comment only makes sense in the context of what was changed. Can this be something like:
```
Original equation  1.1_rkind*heightCanopyTopAboveSnow*log(1._rkind + funcLAI**0.25_rkind) implemented using sqrt() for speed
```

build/source/engine/vegNrgFlux.f90
```diff
@@ -2070,104 +1760,92 @@ subroutine aeroResist(&
...
     if(funcLAI < 0.2_rkind)then
-     z0Canopy = z0Ground + 0.3_rkind*heightCanopyTopAboveSnow*funcLAI**0.5_rkind
+     z0Canopy = z0Ground + 0.3_rkind*heightCanopyTopAboveSnow*sqrt(funcLAI) ! using sqrt intrinsic for speed
```
wknoben, line 1837: As above

build/source/engine/vegNrgFlux.f90
```diff
@@ -3172,10 +2850,11 @@ subroutine aStability(&
  ! ***** process unstable cases
  if(RiBulk<0._rkind)then
   ! compute surface-atmosphere exchange coefficient (-)
-  stabilityCorrection = (1._rkind - 16._rkind*RiBulk)**0.5_rkind
+  stabilityCorrection = sqrt(1._rkind - 16._rkind*RiBulk) ! use sqrt intrinsic for speed
```
wknoben, line 2752: As above

andywood:
I think it's fine to retain some old comments and testing blocks unless
something is absolutely and cleanly finalized and never needs to be seen
again.  We're not developing software for IBM, but rather hydrology models,
and it's not terrible for the codes to keep some of the personality and
thought process of their creators & developers.  E.g., at some point
someone wiped out Martyn's 'dirs_und_philes' subroutine in place of
'dirs_and_files' ... was that really necessary?  Good to clean up code, fix
bugs, improve documentation, remove clutter ... but I'm not sure there's a
need for complete sanitization.

The rkind/i4b issue -- there was a search/replace when switching the model
from fixed types to rkind, and it's reasonable that there are places where
a fixed type makes sense.  indexes are one of those cases (versus state
variables).  probably needs a case by case review when inconsistencies are
found.

build/source/engine/bigAquifer.f90
seantrim: Both `aquiferBaseflowExp - 1_i4b` and  `aquiferBaseflowExp - 1._rkind)` will produce the same numerical value because of Fortran's mixed mode arithmetic rules (e.g., https://fortranwiki.org/fortran/show/Arithmetic or https://docs.oracle.com/cd/E19957-01/805-4939/z400073a2265/index.html). Given that `aquiferBaseflowExp` is of type `rkind`, the `1_i4b` integer literal is automatically promoted to type `rkind` (which has the most storage units of the two data types).

However, after some thought, it seemed  most logical to preserve the original  `aquiferBaseflowExp - 1._rkind)` form because that correctly implies that `aquiferBaseflowExp` is of type `rkind`. I thought that this form would be clearer for developers, hence the revision.

build/source/engine/snowLiqFlx.f90
seantrim: This revision was made for similar reasoning as above. Due to Fortran's mixed mode rules, `mw_exp<1_i4b` and `mw_exp<1._rkind`  will have identical values. However, the original version (`mw_exp<1._rkind`) correctly implies that `mw_exp` is of type `rkind`, which may be clearer for developers.

build/source/engine/soilLiqFlx.f90
seantrim: I have reverted things to two separate lines for easy reading.

build/source/engine/soilLiqFlx.f90
seantrim: I have divided this into separate lines for easy reading.

build/source/engine/vegNrgFlux.f90
seantrim: In this case, `scaleLAI` is of type `rkind`, so `(1_i4b - scaleLAI)` and `(1._rkind - scaleLAI)` give identical values due to Fortran's mixed mode rules (the `1_i4b` literal would get promoted to `rkind` type).

Hypothetically, if `scaleLAI` were of type `i4b`, we may gain accuracy by using `(1_i4b - scaleLAI)` instead.

build/source/engine/vegNrgFlux.f90
seantrim: I have added the improved comment.

build/source/engine/vegNrgFlux.f90
seantrim: I have expanded the comment to give the original equation for context.

build/source/engine/vegNrgFlux.f90
seantrim: I have added the original equation in the comment for context.

build/source/engine/vegNrgFlux.f90
seantrim: The original equation has been added to the comment for context.

build/source/engine/groundwatr.f90
seantrim: It was my understanding that the numerical Jacobian was previously used to confirm the implementation of the analytic Jacobian (which is actually used in the models), and that the numerical Jacobian was no longer used. The fact that `doNumericalJacobian=.false.`  (line 147 in the original file) and cannot be changed without modifying the source code and recompiling is consistent with this. Perhaps @martynpclark can confirm?

build/source/engine/vegNrgFlux.f90
seantrim: If it is deemed acceptable to permit removal of the debug statements, then we can remove the declaration and initialization of the `printflag` logical. I have removed it for now, to be consistent with the rest of the develop_refactor branch. However, if it is decided that the debug statements should be retained, I will restore the logical.

build/source/engine/groundwatr.f90
seantrim: I've confirmed this with @martynpclark during a meeting on June 6th.

build/source/engine/soilLiqFlx.f90
seantrim: I've confirmed removal of this testing code that is no longer used with @martynpclark in a meeting on June 6th.

build/source/engine/groundwatr.f90
seantrim: This change was confirmed with @martynpclark in a meeting on June 6th.

build/source/engine/groundwatr.f90
seantrim: The removal of this debugging statement was confirmed with @martynpclark in a meeting on June 6th because it is no longer used.

seantrim:
Thanks Wouter for looking things over! That exfiltration check in groundwatr.f90 was previously used for debugging purposes but is no longer needed (confirmed with @martynpclark during a meeting). I have amended the whats-new.md file with a statement explaining that several print statements previously used for debugging were removed from the flux files because they were no longer required.

wknoben:
Thanks for the updates @seantrim! 

@martynpclark @andywood  I'll leave this open for a few more days to give you a chance to comment if you want. If not I'll merge this some time next week.

PR #531, opened 2023-07-05
updated whats-new file for v3.2.0 release

wknoben:

- [ ] Describe the change in the release notes (use either `./summa/docs/whats-new.md` or `./summa/docs/minor-changes.md` depending on what changed)

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -2,16 +2,8 @@
 
 This page provides simple, high-level documentation about what has changed in each new release of SUMMA.
 
-## Version 3.0.4 (pre-release)
-- Initial addition of the "What's new" page
-- Added pull request template
-- Adds HRU/GRU info to error messages
-- Fixes a segfault of mysterious origin when using JRDN snow layering
-- Fixes a water balance error w.r.t transpiration
-- Fixes the output message to report the correct solution type
-- Adds tolerance to balance check in updatState.f90
-- Changes all float data types to `rk`, for "real kind", which is intended to make it easier to switch from double to single precision
-- Computes wall clock time for each HRU and time step
+## Version 3.2.0
+- Addition to compute wall clock time for each HRU and time step
 - Fixes a bug that incorrectly writes scalarTotalET and scalarNetRadiation to output in cases where canopy calculations are skipped
 - Canopy ice content check in check_icond.f90 now generates a warning if ice > 0 for T > 0 instead of a graceful exit. Graceful exit still exists if ice > 1E-3.
 - Added case_study folder and Reynolds Mountain East albedo decay experiment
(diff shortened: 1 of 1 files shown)


docs/whats-new.md
andywood, line 1: Can the following be turned off if (a) it adds much time to the run and (b) isn't of interest to the user?  
_## Version 3.2.0
- Addition to compute wall clock time for each HRU and time step_

andywood: This seems like a debugging/development utility feature.  For my westwide HUC12 model it would add ~9B extra calculations (& more operations) to a half-century run, and more if it's discretized.  Perhaps it's trivial -- we may have discussed this previously.

wknoben: We looked at this before when it was implemented in #465. Copying the relevant parts:

> According to [this](https://stackoverflow.com/questions/22122494/is-there-a-cpu-time-performance-penalty-in-fortran), execution time for cpu_time() is not negligible for very large numbers of computations.
> 
> Some quick (online Fortran compiler) testing shows that:
> 
> ```
> program cputimer
>   implicit none
>   integer :: i
>   integer, parameter :: n = 1000000
>   real(8) :: t1, t2, t
>   call cpu_time(t1)
>   do i = 1, n
>     call cpu_time(t)
>   end do
>   call cpu_time(t2)
>   write(*,*) (t2-t1)/n
> end program
> ```
> 
> has an average execution time per iteration of `1.099E-6 s`. If this translates directly from this online tool to regular compiled Fortran code, we're looking at an extra `50 * 365.25 * 8 * 100000 * 1.099E-6 ~= 4.5 hr` of total time added to these CONUS runs. On a per-HRU basis this is not excessive though, at approx. `0.16 s` per HRU
> 
> I also looked at the time cost of an IF-statement like so:
> 
> ```
> program iftimer
>   implicit none
>   integer :: i
>   integer, parameter :: n = 1000000
>   logical, parameter :: calcTime = .FALSE.
>   real(8) :: t1, t2, t
>   call cpu_time(t1)
>   do i = 1, n
>     IF (calcTime) THEN
>       call cpu_time(t)
>     END IF
>   end do
>   call cpu_time(t2)
>   write(*,*) (t2-t1)/n
> end program
> ```
> 
> This gives an average execution time per loop iteration of `2.032E-9 s`, meaning approximately `30 s` added to the CONUS run (approx. `0.0003 s` per HRU).
> 
> On a per-HRU basis either approach is pretty much negligible but for very large domains these calls may start to add up. As far as I can see there are two conflicting principles:
> 
>     * On the one hand, only executing computations when required makes sense;
> 
>     * On the other hand, the code may be kept simpler if we don't excessively use if statements and flags.
> 
> 
> I'm not sure what to suggest but figured I'd share the analysis at any rate to put some (estimated) numbers to this problem.



> Great analysis, Wouter -- thanks for digging into this. Given that we mostly run our models in big split-domain (or some kind of parallelization) modes, I don't think it's a big enough hit to object to adding it, and without the if() statement. For instance, for a 20-node run of a CONUS domain it might end up adding 20-30s, maybe 2 min at much higher resolution. Happy with what others think.

andywood:
Hi Wouter, thanks for digging up the conversation -- it's good that we were
able to land on including it based on that analysis.

wknoben:
No problem :)

PR #539, opened 2023-07-19
with change in scalarthroughfallRain to matLayer state, need to make …

ashleymedin:
The master and develop that are tagged with 3.2.0 are giving SWE balance errors. This fixes it in my tests (and I am currently testing N.Amer). @wknoben


ashleymedin:
To clarify, this PR only affects the calculations to check that SWE balance in an outside loop from the solver. It does not change any of the output variables. The effective rainfall used in the SWE balance, `effRainfall` needs to be averaged only over conditions with a snow layer, and the previous code could average over non-snow situations if the data window was divided (in the case of a step failure), i.e. an originally 3600s step needed to be done in 1800s and 1800s, and the snow was melted for the second 1800s step, then that rainfall should not be included in the effective rainfall on snow over the 3600s.

wknoben:
Seems good to me.

PR #559, opened 2024-02-19
Latest updates to develop-sundials that include enthalpy options

ashleymedin:
This update includes refactor and clean up of the opSplittin routines from Sean, the option to have `nrgConserv = closedForm` (traditional method) or `nrgConserv = enthalpyFD` or `nrgConserv = enthalpyFDlu`, the enthalpy method to conserve energy using the soil enthalpy from the analytical solution or the lookup tables, respectively. 
New output variables that give balance errors can be computed by adding to the outputControl.txt 
```
balanceCasNrg         | 1
balanceVegNrg         | 1
balanceSnowNrg        | 1
balanceSoilNrg        | 1
balanceVegMass        | 1
balanceSnowMass       | 1
balanceSoilMass       | 1
balanceAqMass         | 1
```
Other notable changes are the constraints scale the linesearch in `numMethod = numrec` or `numMethod = kinsol` only on the state variable they are violating, not the entire vector, and the feasiblity/constraints have been corrected for previous bugs in their values. Two end member cases that led to decisions being based on unintialized values (one in the SummaBE solve loop in summaSolv, one in opSplittin) have been fixed, as well as 6 cases of allocation type errors that caused gcc13 to have an internal compiler segfault.


seantrim:
Hi @ashleymedin and @wknoben - I have run the synthetic (aka laugh) and field data (aka WRR) tests for both CH-Earth:develop_sundials and ashleymedin:develop to check for differences. I've used the classic numerical recipes BE solver with the original model decisions to confirm backwards compatibility. 

The compute time and memory usage for both branches are quite similar. I ran the ashleymedin:develop branch second, so the processor may have been slightly less efficient due to temperature. Summaries of computational resource use for each branch can be found in [resources_CH_develop_sundials_PR559.txt](https://github.com/CH-Earth/summa/files/14417186/resources_CH_develop_sundials_PR559.txt) and [resources_ashleymedin_develop_PR559.txt](https://github.com/CH-Earth/summa/files/14417195/resources_ashleymedin_develop_PR559.txt)

There were changes in output for many of the tests. However, differences appear to be small, possibly occurring early in the simulations and accumulating a bit over time. Slightly larger differences occur if the number of model layers changes at different times for each branch. A summary of max differences for all tests can be found in  [output_diffs_PR559.txt](https://github.com/CH-Earth/summa/files/14417166/output_diffs_PR559.txt).

I don't think the output differences are from the refactoring commits in this PR. @ashleymedin , do you have commits with intended output differences in this PR? Perhaps a bug fix?

ashleymedin:
Yes, changing the way the constraints are handled would be enough to make those changes.

wknoben:
Merging as per Ashley's request.

PR #569, opened 2024-06-19
Add CMake Functionality to Summa

KyleKlenk:
I have added the ability use CMake to compile SUMMA. 

I have used two CMakeLists.txt files to achieve this as it helps with Summa-Actors, and follows what the sundials work is using. 

The top level file in `build/` has the configuration and compilation steps. The lower level CMakeLists.txt file inside build/source, just has the files that are required for compilation. This allows Summa-Actors to just import this lower level CMakeLists.txt file instead of having to keep track of the files manually.

The way I have set this up is that inside `build/` is a `cmake` directory. In this directory are a couple helper files for helping CMake find NetCDF and OpenBLAS. LAPACK is the first library that is tried, and if it cannot be found openBLAS is used as a fallback.  I can change this if one is preferred over the other. In addition to the Find files, is a script that can be used to compile Summa in one step `compile_script.sh`. This will configure and build Summa in parallel, all build files will be stored in `cmake_build` which is created by cmake. This `cmake_build` directory has also been added to the `.gitignore` file so no build files are commited.

Lastly, I updated the documentation files, Installation.md specifically to include instructions for using camke

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ Makefile-*
 *.backup
 # site directory from mkdocs
 site/
+# Compilation Files
+build/cmake/cmake_build
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -1,0 +1,104 @@
+cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
+project(summa VERSION 3.2.0 LANGUAGES Fortran)
+
+# Add options for build type
+set(CMAKE_CONFIGURATION_TYPES Release Debug)
+
+# Set Default Executable Name
+set(EXEC_NAME summa.exe)
+set(F_MASTER "${CMAKE_CURRENT_SOURCE_DIR}/..")
+set(EXEC_DIR "${F_MASTER}/bin")
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${EXEC_DIR})
+
+# Use Find files for NetCDF and OpenBLAS
+LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/")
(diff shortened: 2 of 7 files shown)


wknoben:
Looks good to me, thanks! 

Just to be 100% clear for current users, this does not replace the traditional makefiles, nor does it interfere with how they work. This just adds the capability to build using cmake instead, which should be much easier for most users. @KyleKlenk Can you confirm this?

KyleKlenk:
This does not replace the Makefile. I left the original Makefile untouched and users can use it as they normally would.

wknoben:
That's what I thought, thanks for confirming. Looks good to me.

wknoben:
@andywood @martynpclark It's not very conventional but I think merging this straight into `master` makes sense. It might make life easier for current/new users and doesn't affect science results. Let me know if you disagree.

andywood:
Hi Wouter -- seems reasonable especially if you expect a pressing demand
for this feature.  It can be propagated into develop as well. Btw, there
was a comment earlier about cmake being 'easier' than a regular Makefile,
which I wondered about.  I use both and don't find one notably easier than
another.  CMake adds a layer of abstraction.  For the vanilla
summa-on-local-linux use case, neither are difficult and they both work.
Cheers, Andy

wknoben:
Hi Andy, thanks for letting me know. It seems to me that Kyle build in a bunch of features to help locating library paths etc. In my (admittedly somewhat limited) experience it's these kind of things that are easy once you know what you're doing but can be hard to grasp if you're new to compiling code. 

Given that both approaches can exist side-by-side, I think we should just give it a go and readjust if we get negative feedback for some reason.

andywood:
Absolutely!

PR #575, opened 2024-10-08
Update modelwrite.f90 to add in hruId to restart files

### LINKED ISSUE #253: Error when using restart files
When trying to run SUMMA using the initial conditions as a restart file from a previous run I get an error:
```
[andrbenn@n0561 06305-06320]$ /civil/hydro/arbennett/summa/bin/summa_debug -m summa_fileManager.txt -g 6305 16 -p m -r y                                                              
file_master is 'summa_fileManager.txt'.
 GRU-Parallelization run activated. 16 GRUs are selected for simulation.
start at 14:09:13
Name of Model Output control file: /civil/hydro/michaelou/summaProj/summaData/blivneh/summa_run/oc_columbia_v2.txt
forrtl: severe (408): fort: (2): Subscript #1 of the array SNOWDATA has value 6305 which is greater than the upper bound of 16

Image              PC                Routine            Line        Source             
summa_debug        0000000000A5B166  Unknown               Unknown  Unknown
summa_debug        000000000089FA33  read_icond_module          95  read_icond.f90

Stack trace terminated abnormally.
```

I believe (but have not yet confirmed) that this is due to using the GRU Parallelization mode (ie with the `-g 06305 16` option).

DaveCasson:
The need for the hruId index has been discussed in several issues. Notably https://github.com/CH-Earth/summa/issues/253 Also https://github.com/CH-Earth/summa/pull/419, https://github.com/CH-Earth/summa/issues/401

This PR adds writing of hruId to the restart file without modification to other code and while using the existing hruInfo structure.

- [ x] Closes #253 


wknoben:
Nice, thanks! Couple of things for posterity:
- Is there a reason why we cannot easily keep GRU IDs with the same approach? 
- Will this break any existing code that relies on restart files?
- Do we need any tests to confirm the the HRU IDs come out in the correct order?

DaveCasson:
1) GRUs are not currently used at all in the state file, but in testing it appears that there is rationale that GRUs (and gruIds) may need to be included. That is, you can merge currently a number of distributed state files, but there order of the hruIds is not guaranteed to match that the attributes (for example). I'm not sure if adding grus would alleviate this issue or not.  It is left as the responsibility of the modeller (and a difficult mistake to track without hruIds).

2) I don't foresee a case this would break existing code, as it is an additional variable only. Adding GRUs could add more risk?  I do not know.

3)The hru ids being assigned in the correct order comes down to this loop:

 ! Allocate and fill hruIds array
 allocate(hruIds(nHRU))
 do iGRU = 1, nGRU
     do iHRU = 1, gru_struc(iGRU)%hruCount
         cHRU = gru_struc(iGRU)%hruInfo(iHRU)%hru_ix
         hruIds(cHRU) = int(gru_struc(iGRU)%hruInfo(iHRU)%hru_id, kind=i4b)
     end do
 end do
 
So it is a direct pull and write from that existing data structure.  A unit test to compare say to attribute hruId could be useful, but support would be needed to see if / why / how that would be implemented.

ashleymedin:
I have all this in the develop_sundials branch, so you may want to close @wknoben. hruId and gruId are written in the restarts with a call to write_hru_info, https://github.com/CH-Earth/summa/blob/develop_sundials/build/source/netcdf/modelwrite.f90, and these are checked if available in https://github.com/CH-Earth/summa/blob/develop_sundials/build/source/netcdf/read_icond.f90 or assumed the same as the attributes file if not there for backwards compatibility.

wknoben:
ok great, closing for the moment

PR #576, opened 2024-10-10
Sundials merge, continue

ashleymedin:
I put a version of the restart fix that Dave asked for (also has gruId as well as hruId). 
The cmake build is now more portable with openBLAS default and multithreading turned off inside the code.


ashleymedin:
@wknoben, this is a merge to include the restart file upgrade as requested.

wknoben:
Merging as per Ashely's request

PR #578, opened 2025-01-10
Fixed Rootingdepth

ashleymedin:
Bug from Ignacio, just an oversight that rooting depth could be set larger than soil depth, and then throws a Nan.


wknoben:
merging as requested

PR #579, opened 2025-01-28
Continuation of sundials branch merge

ashleymedin:
This set of fixes 

1. correct nan value that results from a large ground roughness parameter
2. splitting to a scalar solution some of the soil scalarv variables were not computed correctly and can lead to balance errors
3. make fluxes output as mean over data window, agreed that this makes more sense than the backwards compatible where they were just the ending solution of the flux (rate over the last substep).


wknoben:
merging as requested

PR #588, opened 2025-05-26
Added maximum infiltration rate decision

wknoben:

- [x] New tests added (describe which tests were performed to test the changes)
- [x] Science test figures (add figures to PR comment and describe the tests)

**Note**
Leaving text below as original, but note that default decision got changed to `topmodel_ga` in this commit: https://github.com/CH-Earth/summa/commit/dea20e6ebe2c2d2a9afb09bdbf00bbb8299ff665

Reason: accurate backward compatibility, no results change for older setups using new code.

--

This PR adds a new decision to SUMMA that allows users to specify the method by which variable `xMaxInfilRate` is calculated. Previously this was Green-Ampt-like, and this adds the original Green-Ampt as an option. Previous science results can be kept identical by setting the new decision to use the old equation. Using the new equation will affect science results. See below.

**Overview**

| Decision name | Option name | Description |
|-|-|-|
| infRateMax | GreenAmpt | (Default) Green-Ampt approximation of infiltration rate. Infiltration rate converges to Ks at full saturation. This option will lead to new science results. |
| | topmodel_GA | Green-Ampt modified to work with a non-constant hydraulic conductivity profile. Infiltration rate converges to 0 for `zScale_TOPMODEL` (SUMMA parameter) values > 1; to something close(-ish) to Ks for `zScaleTOPMODEL = 1`; and to something (much) larger than Ks for `zScale_TOPMODEL < 1`.  This is the only option currently implemented in SUMMA. |
| | noInfExc | Sets the maximum infiltration rate to something veryBig, so that we'll never see a rainfall rate that exceeds it and thus effectively have no infiltration excess |

**Reasoning**

- Previously the maximum infiltration rate always used what's now called the `topmodel_GA` approach. The drawback of this is that in typical scenarios (i.e. `zScaleTOPMODEL >= 1`) maximum infiltration rates are at best equal but typically lower than what Green-Ampt would predict (particularly at higher relative saturation of the soil column) and this leads to relatively more surface runoff than what Green-Ampt would give. 
- Green-Ampt is selected as the default, because this is (likely) the more appropriate choice in most typical SUMMA setups (i.e., those with `noXplicit` or `bigBuckt` groundwater - only `qBaseTOPMODEL` should use the `topmodel_GA` infiltration approach). 

**Changes**

- Added a new model decision to support selection of maximum infiltration rate method
- Added a check to require groundwater option `qBaseTOPMODEL` to be used with infiltration method `topmodel_GA`
- Added a warning that combining groundwater decision `bigBucket` with infiltration method `topmodel_GA` is not recommended, but because what's now `topmodel_GA` was SUMMA's behavior before this change I did not want to make it flat out impossible to use older SUMMA setups. Therefore a warning and not an exit.

**Implementation tests**
The standard test cases are not particularly instructive for these changes because (1) we know they won't be identical because we're implementing something new in the physics routine, and (2) because the changes interact heavily with other other soil parameters and the standard test cases do not showcase this behavior as much as needed to understand this particular change. Therefore I used a different test case: real data from an actual catchment where I systematically looped through all `soilTypIndex` values to ensure the tests cover a wide range of soil settings. Details below.

| Test | Test part 1 | Test part 2 | Expected outcome |
| - | - | - | - |
| 1 | `develop_sundials` before PR | New code with `topmodel_GA` active | Identical simulations: meant to test that adding the decision infrastructure did not change code functioning |
| 2 | `develop_sundials` before PR | New code with `GreenAmpt` active | Different simulations: meant to show order of magnitude impact of new max infiltration rate estimation method |
| 3 | `develop_sundials` before PR | New code with `noInfExc` active | Different simulations: meant to show order of magnitude impact of new max infiltration rate estimation method |
| 4 | New code with `qBaseTOPMODEL` and `GreenAmpt` active | n/a | Graceful exit with message that these decisions cannot be combined |
| 5 | New code with `bigBucket` and `topmodel_GA` active | n/a | Warning that combining these decisions is not recommended but run continues | 

**Test 1**
This test is intended to show that after implementing these changes SUMMA can still generate the same results as before. This is the case: 
- First set of plots shows baseline simulations for the 12 soil types, using `bigBucket` for groundwater.
- Second set of plots shows the new simulations with `bigBucket` coupled to the new `topmodel_GA` decision.
- Third set of plots shows that the differences between both are 0.

test1_outcomes

**Test 2**
This test is intended to show the changes in general hydrograph shape and order of magnitude of differences in `averageRoutedRunoff` simulations, as a result of using the new `GreenAmpt` decision. These are, in most cases, substantial:
- First set of plots shows baseline simulations for the 12 soil types, using `bigBucket` for groundwater.
- Second set of plots shows the new simulations with `bigBucket` coupled to the new `GreenAmpt` decision. `averageRoutedRunoff` is considerably less spiky than in the baseline simulations, as a result of the generally higher maximum infiltration rates estimated by the Green-Ampt method.
- Third set of plots shows that the differences in `averageRoutedRunoff` magnitudes are in the order of `1e-7` for most soil types, which is a comparable order of magnitude to `averageRoutedRunoff` simulated by the baseline configuration.

test2_outcomes

**Test 3**
This test is intended to show the changes in general hydrograph shape and order of magnitude of differences in `averageRoutedRunoff` simulations, as a result of using the new `noInfExc` decision These are, in most cases, substantial:
- First set of plots shows baseline simulations for the 12 soil types, using `bigBucket` for groundwater.
- Second set of plots shows the new simulations with `bigBucket` coupled to the new `noInfExc` decision. `averageRoutedRunoff` is considerably less spiky than in the baseline simulations, because we effectively disable infiltration excess runoff in this case.
- Third set of plots shows that the differences in `averageRoutedRunoff` magnitudes are in the order of `1e-7` for most soil types, which is a comparable order of magnitude to `averageRoutedRunoff` simulated by the baseline configuration.

test3_outcomes

**Test 4**
Code exits early as expected:
```
file_master is './settings/SUMMA/experiments/infiltrationRate_decision/filemanager_topmodel_error_trial.txt'.
   1 controlVersion: SUMMA_FILE_MANAGER_V3.0.0

  [..]

  13 decisionsFile: modelDecisions_topmodel_error.txt

  [..]

FATAL ERROR: summa_paramSetup/mDecisions/maximum infiltration rate method must be topmodel_GA when using qbaseTopmodel for groundwater, not GreenAmpt
```
**Test 5**
Code generates warning but continues running as expected:
```
file_master is './settings/SUMMA/experiments/infiltrationRate_decision/filemanager_topmodelGA_soilTypeIndex_1.txt'.

[..]

   9 groundwatr: bigBuckt

[..]

  28 infRateMax: topmodel_GA
startTime: iyyy, im, id, ih, imin = 2000  1  1  0  0
finshTime: iyyy, im, id, ih, imin = 2010 12 31 23  0
 DEPRECATION WARNING: Combining groundwater parametrization bigBucket with maximum infiltration rate method topModel_GA is not recommended. This was the default in SUMMA v3.x.x and below, but is not appropriate for this groundwater option. Please use Green-Ampt instead.
 
[..]

 Created output file: ./simulations/experiment_infiltrationRate_decision/topmodelGA_infiltrationRate_decision_soilTypeIndex_1_day.nc
 Created output file: ./simulations/experiment_infiltrationRate_decision/topmodelGA_infiltrationRate_decision_soilTypeIndex_1_timestep.nc
2000  1  2  0  0
2000  1  3  0  0
```

**Left to add**
Derivatives for the new Green-Ampt option in `soilLiqFlx.f90`. Currently set to zero.

### DIFF
--- a/build/source/engine/soilLiqFlx.f90
+++ b/build/source/engine/soilLiqFlx.f90
@@ -1257,13 +1257,13 @@ contains
       ! define the hydraulic conductivity at depth=depthWettingFront (m s-1)
       hydCondWettingFront =  surfaceSatHydCond ! Green-Ampt assumes homogeneous soil, therefore the whole soil column has the same hydraulic conductivity
       ! define the maximum infiltration rate (m s-1)
-      xMaxInfilRate = hydCondWettingFront * (1 + (1._rkind - depthWettingFront/sum(mLayerDepth)) * wettingFrontSuction/depthWettingFront) ! Ks * (1 + (Md) * S/F)
+      xMaxInfilRate = hydCondWettingFront * (1._rkind + (1._rkind - depthWettingFront/sum(mLayerDepth)) * wettingFrontSuction/depthWettingFront) ! Ks * (1 + (Md) * S/F)
       ! define the derivatives
-      dxMaxInfilRate_dWat(:) = 0 ! Temporary
-      dxMaxInfilRate_dTk(:)  = 0 ! Temporary
+      dxMaxInfilRate_dWat(:) = -hydCondWettingFront*wettingFrontSuction*dDepthWettingFront_dWat(:)/depthWettingFront**2_i4b
+      dxMaxInfilRate_dTk(:)  = -hydCondWettingFront*wettingFrontSuction*dDepthWettingFront_dTk(:)/depthWettingFront**2_i4b
     case(noInfiltrationExcess)
       ! define the hydraulic conductivity at depth=depthWettingFront (m s-1)
-      !hydCondWettingFront =  surfaceSatHydCond ! this is not needed, but unsure if not setting this will cause problems
+      !hydCondWettingFront =  surfaceSatHydCond ! this is not needed for this calculation, but keeping it here in case not setting this will cause unanticipated problems down the line
       ! define the maximum infiltration rate (m s-1)
       xMaxInfilRate = veryBig ! If maximum infiltration is very big we'll never have a rainfall rate that exceeds it, so no infiltration excess
       ! define the derivatives
--- a/build/source/engine/soilLiqFlx.f90
+++ b/build/source/engine/soilLiqFlx.f90
@@ -80,7 +80,8 @@ USE mDecisions_module,only:   &
   zeroFlux,                   & ! zero flux
   ! look-up values for the maximum infiltration rate parameterization
   GreenAmpt,                  & ! Green-Ampt parameterization
-  topmodel_GA                   ! Green-Ampt parameterization with conductivity profile from TOPMODEL-ish parameterization  
+  topmodel_GA,                & ! Green-Ampt parameterization with conductivity profile from TOPMODEL-ish parameterization  
+  noInfiltrationExcess          ! no infiltration excess runoff
 
 ! -----------------------------------------------------------------------------------------------------------
 implicit none
@@ -88,6 +89,7 @@ private
 public :: soilLiqFlx
 ! constant parameters
 real(rkind),parameter     :: verySmall=1.e-12_rkind       ! a very small number (used to avoid divide by zero)
+real(rkind),parameter     :: veryBig=1.e+20_rkind         ! a very big number (used to effectively set infiltration excess to zero)
 real(rkind),parameter     :: dx=1.e-8_rkind               ! finite difference increment
 contains
 ! ***************************************************************************************************************
@@ -1216,7 +1218,7 @@ contains
   ! **** Update operations for surfaceFlx: flux condition -- main computations (wetting front and derivatives) ****
   associate(&
    ! input: model control
-   ixInfRateMax => in_surfaceFlx % ixInfRateMax , & ! index defining the maximum infiltration rate method (GreenAmpt, topmodel_GA)
+   ixInfRateMax => in_surfaceFlx % ixInfRateMax , & ! index defining the maximum infiltration rate method (GreenAmpt, topmodel_GA, noInfiltrationExcess)
    ! input: depth of upper-most soil layer (m)
    mLayerDepth  => in_surfaceFlx % mLayerDepth  , & ! depth of upper-most soil layer (m)
    ! input: transmittance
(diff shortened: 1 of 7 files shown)


build/source/engine/soilLiqFlx.f90
```diff
@@ -1229,21 +1236,40 @@ subroutine update_surfaceFlx_liquidFlux_computation_wetting_front
...
+     dPart1(:) = surfaceSatHydCond*(zScale_TOPMODEL - 1._rkind) * ( (1._rkind - depthWettingFront/total_soil_depth)**(zScale_TOPMODEL - 2._rkind) ) * (-dDepthWettingFront_dTk(:))/total_soil_depth
+     dPart2(:) = -dDepthWettingFront_dTk(:)*wettingFrontSuction / (depthWettingFront**2_i4b)
+     dxMaxInfilRate_dTk(:)  = fPart1*dPart2(:) + fPart2*dPart1(:)
+    case(GreenAmpt)
+      ! define the hydraulic conductivity at depth=depthWettingFront (m s-1)
+      hydCondWettingFront =  surfaceSatHydCond ! Green-Ampt assumes homogeneous soil, therefore the whole soil column has the same hydraulic conductivity
+      ! define the maximum infiltration rate (m s-1)
+      xMaxInfilRate = hydCondWettingFront * (1 + (1._rkind - depthWettingFront/sum(mLayerDepth)) * wettingFrontSuction/depthWettingFront) ! Ks * (1 + (Md) * S/F)
+      ! define the derivatives
+      dxMaxInfilRate_dWat(:) = 0 ! Temporary
```
seantrim: `dxMaxInfilRate_dWat(:) = -hydCondWettingFront*wettingFrontSuction*dDepthWettingFront_dWat(:)/depthWettingFront**2_i4b`

build/source/engine/soilLiqFlx.f90
```diff
@@ -1229,21 +1236,40 @@ subroutine update_surfaceFlx_liquidFlux_computation_wetting_front
...
+      dxMaxInfilRate_dTk(:)  = 0 ! Temporary
```
seantrim: `dxMaxInfilRate_dTk(:) = -hydCondWettingFront*wettingFrontSuction*dDepthWettingFront_dTk(:)/depthWettingFront**2_i4b`

seantrim (changes requested):
Hi @wknoben - I have reviewed all changes. All looks correct apart from the derivatives of the max infiltration rate, which were temporarily set to zero. I have added comments to lines 1262-1263 for soilLiqFlx.f90 with suggestions for the derivative formulas.

The `GreenAmpt` derivative formulas were derived by reusing logic from the `topmodel_GA` derivatives. Namely, it is assumed that only the `depthWettingFront` variable depends on the water and energy state variables.

wknoben:
@seantrim Changed, thanks!

martynpclark (approved):
Everything looks great

wknoben:
**Test 6**
Documenting that code executes as expected if a filemanager is provided that does not include the new `infRateMax` decision. "As expected" here means defaulting to using the new Green-Ampt code.

seantrim (approved):
Updates were successfully verified!

wknoben:
For posterity, default decision got changed to `topmodel_ga` in this commit: https://github.com/CH-Earth/summa/commit/dea20e6ebe2c2d2a9afb09bdbf00bbb8299ff665

PR #594, opened 2025-06-20
Fix snow sublimation bug

moghairib:

- [x] New tests added (describe which tests were performed to test the changes)
- [x] Science test figures (add figures to PR comment and describe the tests)

# Summary of the problem

This PR is fix a bug in the snow sublimation calculations related to the wind profile. Especifically, the bug is noted when the wind profile is transitioned from exponential to log below canopy.
This bug was identified during the SUMMA summer madness when a test case showed substantial snow sublimation for the default model configuration/parameter set. The test case uses `MODIFIED_IGBP_MODIS_NOAH` vegetation lookup table for a `woody savanna` case.

The suggested changes were to ensure that `heightCanopyBottomAboveSnow >= z0Ground + xTolerance` instead of `heightCanopyBottomAboveSnow >= 0` as this caused the potential wind profile to continue to the ground surface (i.e., removing the log profile under canopy).

The change was suggested and validated by @martynpclark.

The modified code resulted in a more realistic snow sublimation compared to the unmodified code (see below for the same default parameter set).
original
aftermod

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,5 @@ build/cmake_build
 runinfo.txt
 # OpenWQ source code
 build/source/openwq/openwq
+# vscode related files
+.vscode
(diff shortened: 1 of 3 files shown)


moghairib:
@martynpclark, please review this.

martynpclark (approved):

Issue #603, opened 2025-08-08
Add something to view model decision options without diving into `mDecisions.f90`


wknoben:
Issue: figuring out the exact spelling for model decisions is a bit cumbersome right now (requires knowing that these are found in `mDecision.f90` for one, and then opening the file and finding the correct lines).

Some command line argument like:
```
> summa.exe --decision-options hc_profile
Current options for hc_profile (choice of hydraulic conductivity profile):
- constant: constant hydraulic conductivity with depth
- pow_prof: power-law profile
``` 
could be helpful. Unsure how feasible this is though. 

Alternatively, clarifying various error messages could also help. E.g.:

```
FATAL ERROR: summa_paramSetup/mDecisions/power-law hydraulic conductivity profile must be selected 
when using topmodel baseflow option
```
would be better as:
```
FATAL ERROR: summa_paramSetup/mDecisions/power-law hydraulic conductivity profile must be selected 
when using topmodel baseflow option. Set 'hc_profile' to 'pow_prof' in model decisions input file.
```

This would also help users fix "unknown decision" errors more quickly. E.g.
```
FATAL ERROR: summa_paramSetup/mDecisions/unknown hydraulic conductivity profile [option=power-law]
```
would be easier as:
```
FATAL ERROR: summa_paramSetup/mDecisions/unknown hydraulic conductivity profile [current option=power-law; available = constant, pow_prof]
```
but this would require keeping the list of decisions updated in all these different statements as well. Reading this programmatically from somewhere would make maintenance easier.

PR #604, opened 2025-08-08
clarified model decision error messages with exact wording needed in mDecisions file

wknoben:
Couple of quick improvements towards #603.

Error message are easy to address (compared to a general solution) because these require very specific combinations of model decisions that are unlikely to change.

Tested by modifying an mDecisions file and triggering each message. Code compiles and errors/warnings as expected.


wknoben:
@ashleymedin could you give this a quick glance to check I didn't miss something? Thanks!

ashleymedin:
This looks good to me.

PR #605, opened 2025-08-09
Change MVT and NLUS from 200 to 300

ijaguirre:

- [ ] Describe the change in the release notes (use  `./summa/docs/whats-new.md`)

I'd like to suggest a pull request change MVT and NLUS from 200 to 300 to allow more flux towers to run without changes in the code. 
This change does not change the science.


wknoben:
This is probably fine as, but is there a specific reason for setting it to 300 instead of, say, 10,000? Adding a sentence or so to this PR discussion about how long we expect 300 to be sufficient for could be good.

More generally, I believe this is needed to read tower-specific input tables right? Does this point to a more general need to overhaul the table structure so switching between 1 lookup table everywhere to site-specific lookup tables is easier? Might be good to open an issue for this to note it as a future enhancement.

wknoben:
@ijaguirre just checking if you saw the message above

ijaguirre:
You’re right. The reason I initially suggested 200/300 instead of 10,000 is because my configuration currently includes around 200 flux towers. Setting the value to 10,000 would indeed address the problem more broadly and provide a long-term solution.

A more comprehensive approach, however, would be to integrate LAI time series directly into the trial parameters. This would bring additional benefits, such as the ability to represent land cover changes, disturbances, and fires—factors that cannot be fully captured using only winter_lai, summer_lai, or a static LAI table. With this in mind, I suggest evaluating an alternative way of incorporating LAI into the model.

In the meantime, I’m happy to adjust the value to 1,000 or 10,000 if you feel that would be more appropriate.

PR #606, opened 2025-08-11
Bugfix: Smax calculations in FUSE ARNO/VIC and PRMS saturation excess options

wknoben:
Smax needs to be storage in the soil column (layer depth * porosity), not just soil column depth.

Compiles as expected, and a simple test case (1 month, hourly sims) runs to completion with either modeling decision activated.


wknoben:
@seantrim - could you have a look please?

seantrim (approved):
The changes look good - thanks @wknoben for improving the max water storage calculations.

Issue #610, opened 2025-10-26
create capabilities to read buffered input data


martynpclark:
There are too many input operations for cases of calibrating lumped summa simulations across multiple catchments. In this case one data point is read at a time for (potentially) thousands of summa instances -- and these data reads all occur simultaneously. This gives far too many IOPS (Input/Output Operations Per Second).

There is a need to add capabilities to read data from the entire forcing file simultaneously -- for 11 years of hourly data in a forcing file, this can reduce IOPS 100k times.

ashleymedin:
We have this in develop_sundials now, @wknoben. You can see it here https://github.com/CH-Earth/summa/blob/develop_sundials/build/source/driver/summa_writeOutput.f90, and it merged/updated the code that @martynpclark started.

wknoben:
great, thanks

PR #611, opened 2025-10-26
include capabilities for buffered read of input data

martynpclark:
Closes issue #610

Code provides bit-for-bit matching for the Reynolds Mountain East test case. More testing is needed.

Seeking review and testing managed by @wknoben


wknoben:
to do:
- multiple input files
- multiple output files
- multiple GRUs
- multiple HRUs
- with `-g` option

Issue #609, opened 2025-10-25
update summa documentation


martynpclark:
There are a few updates required for the summa documentation for the `summa_sundials` branch. These updates will be required as we move to summa_v4.

1. Consider if the Makefile instructions are still necessary of if we want to use Cmake exclusively.

2. The `Cmake` instructions point to a compile script that does not exist. Provide a list of scripts in this directory that can be used with short descriptions of each script.

3. The description of dependencies is quite long. Suggest a short bulleted list with links to longer descriptions.

4. There is a need to install sundials. Here is my approach:

```
git clone https://github.com/LLNL/sundials.git
cd sundials
mkdir build && cd build
cmake ..   -DCMAKE_INSTALL_PREFIX=$HOME/local/sundials   -DBUILD_SHARED_LIBS=ON   -DBUILD_IDA=ON -DBUILD_IDAS=ON   -DBUILD_FORTRAN_MODULE_INTERFACE=ON
make
make install
```

It is important to note that the fortran module interface should be on (it is off by default)

5. The test cases are now on HydroShare. Suggest point directly to the hydroShare test cases.

wknoben:
@martynpclark  There are differences between the `master` and `develop` (as well as `develop_sundials`) branch. Do we want this to go into `master` or split out the docs for `develop_sundials` and merge again when we integrate V4 into `master`?

wknoben:
all fixed in #612

Issue #607, opened 2025-08-11
Pre-check for NAN


ijaguirre:
Hi, 

I'd like to suggest that SUMMA could check the presence of NANs before running the entire code. By mistake, I attempted to run a simulation with NANs in the precipitation timeseries, and I got errors related to WETBULB and the following one:

```
rVec =  -0.69395E+01          NaN  0.96080E-06          NaN -0.13730E-06 -0.93862E+03 -0.15351E-08 -0.13028E+02  0.18175E-07  0.32740E+02  0.10888E-07 -0.13346E+02  0.71326E-08 -0.51131E+01  0.68446E-08  0.22056E+00  0.60971E-08  0.15842E+01 -0.12916E-08 -0.35749E-11
fVec =   0.69395E+01          NaN -0.96080E-06          NaN  0.14416E-06  0.93862E+03  0.72738E-08  0.13028E+02 -0.14223E-07 -0.32740E+02 -0.87824E-08  0.13346E+02 -0.63382E-08  0.51131E+01 -0.65914E-08 -0.22056E+00 -0.60312E-08 -0.15842E+01  0.12979E-08  0.35749E-11


FATAL ERROR: run_oneGRU (gru nc = 1, gruId = 1)/run_oneHRU (hru nc = 1, hruId = 1)/coupled_em/opSplittin/varSubstep/systemSolv/summaSolve4ida/eval8summaWithPrime/computResidWithPrime/vector of

[ERROR][rank 0][./summa_develop_sundials/sundials/sundials-7.3.0/src/ida/ida.c:2384][IDAHandleFailure] At t = 0the residual function failed unrecoverably.
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG
STOP 1
```
Following @wknoben's advice, I checked for NANs, fixed the forcing file, and ran the simulations. 
To avoid this problem for future users, evaluating the presence of NAN in any forcing variable would be beneficial.
Thanks

andywood:
This idea has come up before - i.e., the question being, how much work
should summa do to check its own inputs?  It could do all kinds of things,
ie NaN checks, range checks, cross-variable consistency, and so on; and
this kind of QC could be extended to attributes and parameters, initial
states, as well.   The two main cons are probably code bloat and some run
time costs.  If a user has a well established set of NaN-free inputs or a
robust input preparation process, they may not want to model to take time
with that kind of screening.  If it were introduced, I'd suggest that it be
[...]

PR #612, opened 2025-10-30
V4 docs update

wknoben:
Working on #609.  More to be done.


wknoben:
#607 closed by 5b84bef

wknoben:
This should address everything in #609 in at least an initial fashion. I kept the Makefile instructions for the moment but made them less prominent. We can revisit later.

Related changes: this brings all the docs into the `develop_sundials` branch as well. Everything builds locally but I'll still need to see if readthedocs picks it up correctly.

PR #616, opened 2025-12-03
updated docs with instructions on adding variables

wknoben:
Builds on local.


wknoben:
@brancat - can you review please?

PR #624, opened 2026-04-28
Merge for stable release of develop_sundials

ashleymedin:
This has successfully run with Actors and with the buffered write option with the 5 numerical methods of the enthalpy paper plus the reference solution, on Fir, for a 1% of N. America. Results are the same. This can be the stable release as far as I am concerned.


wknoben:
merging as requested

PR #626, opened 2026-05-29
Continuing Sundials merge

ashleymedin:
This fixes some more bugs in the subroutine satHydCond: compute vertical profile of saturated hydraulic conductivity, smoothes infiltration blockage near saturation, fixes baseflow derivatives, and removes the f-Richards decision (optional, backwards compatible, but moisture form is no longer pretending to be supported).


wknoben:
merging as requested @ashleymedin

PR #631, opened 2026-08-23
Feature/domain parallel

martynpclark:
## Description

This PR reimplements the MPI domain-parallelization approach previously developed in [`junwei-guo/summa-mpi`](https://github.com/junwei-guo/summa-mpi) and described by Guo et al. (2026). That implementation distributes GRUs across MPI ranks, with each rank allocating memory for its assigned domain, reading its own forcing data, advancing the model independently, and writing rank-specific output.

The reimplementation was needed because the original MPI development was maintained separately from the main SUMMA repository and had diverged substantially from the current codebase. More importantly, this PR reorganizes the MPI implementation to provide a cleaner separation of concerns and a more flexible foundation for future parallel execution strategies.

MPI-specific initialization and execution are isolated in a new MPI driver, while the core SUMMA model routines remain unaware of MPI. The build system now produces separate serial and MPI executables, allowing the same SUMMA codebase to support both execution modes without introducing MPI dependencies into the serial executable.

The domain decomposition itself remains SUMMA-specific: the run domain is partitioned by GRU, each rank constructs its local GRU/HRU mapping, and model data structures are allocated and populated only for the local domain. This follows the coarse-grained MPI strategy described by Guo et al. (2026), in which GRUs are distributed among ranks and remain in rank-specific memory.

The PR also introduces the beginnings of [`parallel-utils`](https://github.com/CH-Earth/parallel-utils), a deliberately lightweight, model-agnostic set of Fortran utilities for parallel execution. `parallel-utils` is maintained as a separate repository and is included in SUMMA as a **Git submodule**, allowing the parallel utilities to evolve independently of SUMMA and potentially be reused by other models. At present it contains only basic MPI infrastructure, but the intent is to provide reusable primitives that can support a broader range of parallel workflows. In particular, the cleaner separation introduced here is intended to make it easier to add other MPI execution patterns, including model-instance parallelization for ensembles, calibration, and parameter estimation, without entangling those concerns with SUMMA's domain-parallel implementation.

## Domain localization and data access

The localization work also prompted a broader cleanup of SUMMA's input-data routines. A key part of this refactor is that the code now distinguishes explicitly between three spatial index domains:

- **File domain** — the complete set of GRUs and HRUs present in the input files (`nGRU_file`, `nHRU_file`). These indices identify locations in NetCDF input files.
- **Run domain** — the subset of GRUs selected for a particular SUMMA execution (`startGRU_domain`, `nGRU_domain`). For a full-domain simulation this is the complete file domain; for `-g` or `-h` runs it is a selected subset.
- **Local domain** — the subset of the run domain assigned to the current process or MPI rank (`startGRU_local`, `nGRU_local`, `nHRU_local`). For serial execution the local domain is the run domain; for MPI execution each rank operates only on its assigned local subset.

The GRU–HRU mappings are constructed once for the local domain and retain the corresponding input-file indices. This allows subsequent data readers to operate on local SUMMA data structures while explicitly mapping those structures back to the appropriate records in each input file.

This distinction is important because the ordering of HRUs in the initial-condition file is not necessarily the same as the ordering in `LocalAttributes` or the forcing files. The initial-condition reader therefore constructs explicit local-to-file mappings, while forcing reads use the `LocalAttributes` file indices and determine the contiguous HRU range required by each rank.

Together, these changes make the data readers more consistent and robust, clarify the distinction between model-local and file indexing, remove duplicated mapping logic, and provide a common data-access model for both serial and distributed execution.

## Regression testing

This PR adds a small regression-testing suite under `utils/regression/` for validating both the serial and MPI execution paths. The test workflow runs the reference serial executable, the new serial executable, and the MPI executable using 1, 2, 4, 8, 12, and 16 ranks on the same multi-GRU test domain.

The resulting NetCDF files are compared across all model variables after reconstructing the complete domain from the rank-specific MPI output files. These tests give **bit-for-bit identical results (maximum difference = 0)** between the new implementation and the reference serial implementation for every tested MPI decomposition.

The regression tests were performed independently against:

- the SUMMA release tag from which the MPI redevelopment began (`v4.0.0-exp-sundials.2`); and
- the current tip of `develop_sundials`, after rebasing this PR onto that branch.

Both comparisons pass with bit-for-bit agreement. This provides a direct check that the domain localization, input-data refactoring, and MPI domain decomposition do not alter the numerical results.

The regression utilities also include simple plotting tools for visual inspection of selected model variables and timing utilities for comparing parallel performance.

## Performance testing

Basic strong-scaling tests were performed on an Apple M4 Max (16 cores: 12 performance and 4 efficiency; 128 GB memory), using GNU Fortran 16.1.0 and Open MPI 5.0.9 on macOS 26.6. These tests are intended as a development check rather than a formal HPC benchmark. The new implementation shows essentially the same scaling behavior as the original `summa-mpi` implementation, reaching approximately 9x speedup at 12–16 MPI processes. Across the tested process counts, the new implementation performs comparably to, and generally slightly faster than, the original implementation. These tests therefore provide an initial check that the architectural refactoring has not introduced a performance penalty relative to the original MPI implementation, rather than serving as a formal scalability benchmark.

More extensive distributed-memory scaling tests are left for future evaluation. The original MPI implementation has previously demonstrated near-linear scaling to 1,024 cores on HPC systems.

## Summary of changes

- Reimplements the previously developed SUMMA MPI domain-parallelization capability in the current codebase.
- Separates MPI execution from the core SUMMA model through a dedicated MPI driver.
- Builds separate serial and MPI executables.
- Restricts direct MPI dependencies to the parallel execution infrastructure.
- Introduces explicit file, run, and local spatial domains.
- Localizes GRU/HRU allocation and processing to each MPI rank.
- Refactors input-data handling around explicit local-to-file mappings.
- Improves handling of differing HRU orderings among SUMMA input files.
- Adds the lightweight, model-agnostic [`parallel-utils`](https://github.com/CH-Earth/parallel-utils) submodule as the basis for reusable Fortran parallel utilities.
- Adds regression, timing, and visualization utilities for validating parallel execution.
- Demonstrates bit-for-bit agreement with serial SUMMA across all tested MPI decompositions.
- Retains performance comparable to the original `summa-mpi` implementation.

## Reference

Guo, J., Clark, M. P., Knoben, W. J. M., Keshavarz, K., Klenk, K., Van Beusekom, A., et al. (2026). *Implementation and evaluation of parallel computing approaches for large-domain, process-based hydrologic simulations*. Journal of Advances in Modeling Earth Systems, 18, e2025MS005064. https://doi.org/10.1029/2025MS005064


junwei-guo:
Could we add a few simple steps or instructions to the document to help users enable the parallel git submodule in SUMMA-sundial?

junwei-guo (approved):
The code has been tested on DRAC Fir using 192 to 960 CPU cores. The test case covers a North American domain with approximately half a million GRUs. The results demonstrate excellent scaling across the tested core counts. Total memory usage remains largely consistent, with only minor variations. Input/output performance is also excellent, and the CPU core idle rate is approximately 10–15%.

Based on these results, the code is ready for broader public testing and production runs.

One optional improvement concerns console output in MPI applications. Currently, each MPI rank can print messages to the console, which results in many duplicate messages. In the future, console output could be restricted to a designated rank (e.g., rank 0) to reduce redundant output and improve readability.

PR #632, opened 2026-08-27
Feature/mizuroute

martynpclark:
This PR adds optional mizuRoute network routing to SUMMA, establishing the infrastructure needed to configure, initialize, run, and output routed streamflow within the SUMMA simulation workflow. It adds mizuRoute and toml-f as dependencies, introduces TOML-based configuration alongside the existing SUMMA file-manager workflow, and reorganizes the CMake dependencies to keep the mizuroute integration optional while isolating conflicting Fortran module namespaces. The coupling architecture maintains a thin boundary between the models: SUMMA-facing routines translate SUMMA configuration and runoff into the mizuRoute interface, while native mizuRoute initialization and routing continue to operate on mizuRoute data structures.

The coupled workflow runs end-to-end: SUMMA computes basin runoff, transfers the complete runoff field to mizuRoute each timestep, performs network routing, and writes reach-level routed streamflow, upstream area, and routing-method metadata into the existing SUMMA timestep NetCDF output. The output implementation supports both per-timestep and full-series buffering while retaining SUMMA's ownership of the output files. The coupled implementation has been validated against standalone mizuRoute simulations, with routed streamflow matching across all reaches, and regression tests have been added to verify this behavior. The coupled and sequential workflows, coupling design and architecture, build procedures, configuration, and current parallelization limitations are now included in the SUMMA documentation.


junwei-guo:
### Summary

This PR introduces TOML-based input configuration to SUMMA, providing users with a clear, structured way to configure SUMMA and river routing, as well as calibration inputs in the latest branch.

The PR couples SUMMA’s land surface simulations with mizuRoute’s river network routing. The integration includes:

- Initializing mizuRoute during SUMMA initialization.
- Passing GRU-level runoff from SUMMA to mizuRoute after each land surface time step, then advancing river routing by one time step.
- Writing routing results to NetCDF files using SUMMA’s existing output routines.

The coupled implementation was evaluated for the Bow River catchment at Banff, with results consistent with those from the decoupled implementation.

### Optional suggestions

- **Support standalone routing within the coupled framework.** Users may need to run routing repeatedly without rerunning SUMMA’s land surface simulations—for example, when optimizing routing parameters. A routing-only mode could reuse the coupled framework’s configuration and previously generated runoff, avoiding the need to compile and configure a separate mizuRoute executable.

- **Explore parallelization and spatial decomposition.** MPI-based domain parallelization has been implemented and tested in the legacy SUMMA code. Extending and evaluating this capability in the coupled SUMMA–mizuRoute framework would be a useful direction for future work.

### Concern: mizuRoute dependency stability

mizuRoute is included as an external Git submodule from [[ESCOMP/mizuRoute](https://github.com/ESCOMP/mizuRoute.git)](https://github.com/ESCOMP/mizuRoute.git). A Git submodule normally references a specific commit, so upstream changes are not automatically incorporated unless the submodule is explicitly updated. However, if the build or update workflow tracks the latest main-branch version, upstream changes could introduce compatibility issues.

Could we pin the dependency to a tested, stable mizuRoute release or commit and document the update process? This would help ensure compatibility and reproducibility.

martynpclark:
Responding to @junwei-guo 

1. Support standalone routing within the coupled framework. This is supported -- there is a script to build mizuroute here:
https://github.com/CH-Earth/summa/blob/develop/utils/test/test_mizuroute/make_mizuRoute.sh

2. Explore parallelization and spatial decomposition. Yes, parallelization of mizuroute still needs to be implemented. There are warnings in the code here:
https://github.com/CH-Earth/summa/blob/develop/build/source/driver/summa_config.f90#L224-L230

3. Concern: mizuRoute dependency stability. Could we pin the dependency to a tested, stable mizuRoute release or commit and document the update process? This would help ensure compatibility and reproducibility. Yes, the mizuroute submodule is pinned to v3.1.1 so the implementation is not sensitive to changes to the mizuroute code base.

